PR #200, opened 2026-03-18
fix: avoid recovering underfunded session channels

arjunblj:
## Summary
- only recover a suggested session channel when its remaining on-chain balance can cover the next payment
- fall back to opening a new channel instead of issuing a voucher that server-side session validation must reject
- add direct recovery coverage plus a session-level regression test for the fallback path

## Test plan
- [x] `pnpm check:ci`
- [x] `pnpm check:types`
- [x] `pnpm check:compat`


brendanjryan:
closing as this is now stale / fixed in sessions v2

PR #220, opened 2026-03-23
feat: add html payment pages for browser 402 responses

tmm:
Experimental: Renders 402 as html in browsers with way to complete payment (~universal payment links).

### DIFF
--- a/examples/_shared/mppxSource.server.ts
+++ b/examples/_shared/mppxSource.server.ts
@@ -1,0 +1,1 @@
+export { Mppx, Store, stripe, tempo } from '../../src/server/index.js'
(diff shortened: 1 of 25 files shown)


src/server/Mppx.ts
```diff
@@ -233,6 +234,15 @@ export declare namespace create {
     methods extends Methods = Methods,
     transport extends Transport.AnyTransport = Transport.Http,
   > = {
+    /**
+     * Serve an HTML payment page to browsers (requests with `Accept: text/html`).
+     *
+     * - `true` — use the built-in payment page
+     * - `(challenge) => string` — custom HTML renderer
```
brendanjryan: nice

PR #222, opened 2026-03-24
feat: add restored session manager support and harden SSE startup

0xCaso:
## Summary
- add `SessionManager` restore support so callers can resume Tempo session state across process restarts
- export the Tempo client entrypoint and harden SSE startup by preserving `Headers` instances and failing fast on non-OK SSE responses
- keep restored close behavior consistent by clearing opened state after a successful close

## Motivation
- OpenCode-style clients restart frequently, but still need to continue using existing Tempo session channels
- without restore support, callers can only rely on in-memory session state and lose continuity after process restart
- the SSE hardening changes make restored and fresh session flows fail more predictably when the upstream does not actually return a valid stream

## Test Plan
- [x] `VITE_NODE_ENV=testnet pnpm test --project node src/tempo/client/SessionManager.test.ts src/tempo/client/Session.test.ts`
- [x] `pnpm run check:types`<

PS: i left the docs file just for reference

### DIFF
--- a/docs/superpowers/specs/2026-03-22-session-manager-restore-design.md
+++ b/docs/superpowers/specs/2026-03-22-session-manager-restore-design.md
@@ -1,0 +1,219 @@
+# SessionManager Restore API Design
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:writing-plans after this spec is approved to create the implementation plan.
+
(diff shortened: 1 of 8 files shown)


brendanjryan:
all of this is modeled in the new sessions interface! 

https://mpp.dev/blog/sessions-improved

0xCaso:
@brendanjryan sick, gg! that's a big one 🙏

PR #231, opened 2026-03-24
feat(tempo): add split payments for charge

brendanjryan:
## Summary
- add `splits` support to Tempo charge request validation and wire serialization
- build ordered primary-plus-split transfer sequences on the client and verify them for both `transaction` and `hash` credentials on the server
- extend fee-payer validation for split transfers and add split-focused tests plus a changeset

Spec: https://github.com/tempoxyz/mpp-specs/pull/203

PR #232, opened 2026-03-24
fix: bind tempo attribution memo suffix to challenge id

brendanjryan:
## Summary
- bind SDK-generated Tempo attribution memos to the current challenge by reusing the trailing 7 bytes
- add same-challenge replay protection in `Mppx` via `NonceSet` and a reserved internal challenge nonce
- keep explicit custom memos and plain/no-memo transfers on the existing store-backed fallback path

## Testing
- `pnpm check:types`
- `pnpm exec vitest run src/internal/NonceSet.test.ts src/tempo/Attribution.test.ts src/server/Mppx.test.ts`
- `pnpm exec vitest run src/tempo/server/Charge.test.ts`

### DIFF
--- a/src/server/Mppx.ts
+++ b/src/server/Mppx.ts
@@ -296,7 +296,12 @@ function createMethodFn(parameters: createMethodFn.Parameters): createMethodFn.R
         const challenge = Challenge.fromMethod(method, {
           description,
           expires,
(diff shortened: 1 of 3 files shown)


brendanjryan:
depends on https://github.com/wevm/mppx/pull/231

PR #233, opened 2026-03-24
fix: add client-side challenge expiration validation

### LINKED ISSUE #198: UX: Client-side should validate challenge expiration before creating credential
## Problem

When a client receives a Challenge with an expires timestamp that has already passed, the Mppx.createCredential() method still attempts to create a credential and make the payment.

This results in:
1. Wasted user time - client does work that will be rejected
2. Unnecessary network round-trips
3. Potential user confusion (why did my payment fail?)

## Current Behavior

The client does not check challenge expiration before creating credentials.

The server validates expiration (src/server/Mppx.ts:359-367):

```ts
if (credential.challenge.expires && new Date(credential.challenge.expires) < new Date()) {
  return { challenge: response, status: 402 }
}
```

But by then the client has already:
1. Connected wallet
2. Signed a transaction/message
3. Submitted the credential

## Suggested Fix

Add an expiration check in createCredential() before attempting to create the credential:

```ts
async createCredential(response, context) {
  const challenge = transport.getChallenge(response)
  
  // Early rejection of expired challenges
  if (challenge.expires && new Date(challenge.expires) < new Date()) {
    throw new Errors.PaymentExpiredError({ expires: challenge.expires })
  }
  
  // ... rest of implementation
}
```

## Impact

- Type: UX improvement
- Risk: Low - just fails faster with clearer error
- Breaking: No - just changes when the error is thrown

JasonOA888:
## Summary

Fixes #198

## Problem

When a client receives a Challenge with an `expires` timestamp that has already passed, the client still attempts to create a credential and make the payment. This results in:

1. Wasted user time - wallet interactions that will be rejected
2. Unnecessary network round-trips
3. User confusion (why did my payment fail?)

The server already validates expiration (`src/server/Mppx.ts:359-367`), but by then the client has already connected wallet and signed transactions.

## Solution

Adds client-side expiration validation at two code paths:

1. **`Fetch.from()`** - auto-402 interceptor, after challenge matching
2. **`Mppx.createCredential()`** - manual API, after challenge parsing

Both throw `PaymentExpiredError` (already defined in `Errors.ts`) when `challenge.expires` is set and in the past.

## Changes

```ts
// src/client/internal/Fetch.ts - after finding matching challenge
validateChallengeExpiration(challenge)

// src/client/Mppx.ts - in createCredential()
if (challenge.expires && new Date(challenge.expires) < new Date()) {
  throw new Errors.PaymentExpiredError({ expires: challenge.expires })
}
```

## Testing

```ts
// Expired challenge
const challenge = { expires: "2025-01-01T00:00:00Z", ... }
await mppx.createCredential(response)
// Throws: PaymentExpiredError: Payment expired at 2025-01-01T00:00:00Z.
```

## Impact

- Type: UX improvement
- Risk: Low - fails faster with clearer error
- Breaking: No - just changes when the error is thrown

### DIFF
--- a/src/client/Mppx.ts
+++ b/src/client/Mppx.ts
@@ -1,4 +1,5 @@
 import type * as Challenge from '../Challenge.js'
+import * as Errors from '../Errors.js'
 import type * as Method from '../Method.js'
 import type * as z from '../zod.js'
 import * as Fetch from './internal/Fetch.js'
@@ -80,6 +81,11 @@ export function create<
     transport,
     async createCredential(response: Transport.ResponseOf<transport>, context?: unknown) {
       const challenge = transport.getChallenge(response as never) as Challenge.Challenge
+
+      // Validate challenge expiration before creating credential (client-side early rejection)
+      if (challenge.expires && new Date(challenge.expires) < new Date()) {
+        throw new Errors.PaymentExpiredError({ expires: challenge.expires })
+      }
 
       const mi = methods.find((m) => m.name === challenge.method && m.intent === challenge.intent)
       if (!mi)
--- a/src/client/internal/Fetch.ts
+++ b/src/client/internal/Fetch.ts
@@ -1,4 +1,5 @@
 import * as Challenge from '../../Challenge.js'
+import * as Errors from '../../Errors.js'
 import type * as Method from '../../Method.js'
 import type * as z from '../../zod.js'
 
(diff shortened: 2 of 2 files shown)


brendanjryan:
Thanks for suggesting! 

We initially didn't implement this, since the server is authoritative and the client could have clock skew from the server expiry. We will consider this in a new version of the SDK if we see it more frequently

PR #247, opened 2026-03-26
fix(session): separate sender from fee payer in settle and close

brendanjryan:
The escrow contract requires `msg.sender == payee` for `settle()` and `close()`. The `sendFeePayerTx` helper used the fee payer as both sender and gas sponsor, causing every fee-sponsored settlement/close to revert with `NotPayee()`.


brendanjryan:
@tmm examples are good here!

PR #327, opened 2026-04-10
fix: increase `maxTotalFee` to 50b

o-az:
user report

> After upgrading from 0.5.5 to 0.5.10, tempo.charge() with feePayer: true fails on Moderato (testnet) with:
> FeePayerValidationError: fee-sponsored transaction total fee budget exceeds sponsor policy gas: 626497 maxFeePerGas: 24000000000 totalFee: 15035928000000000
> The hardcoded maxTotalFee is 10_000_000_000_000_000 (0.01 ETH). A standard transferWithMemo estimates ~626k gas at 24 gwei, totaling 0.015 - 50% over the cap.
> This worked on 0.5.5 (no policy check existed). Is 626k gas expected for a memo transfer on Tempo? Or is the policy cap too low?

### DIFF
--- a/.changeset/honest-wasps-sniff.md
+++ b/.changeset/honest-wasps-sniff.md
@@ -1,5 +1,5 @@
 ---
-"mppx": patch
+'mppx': patch
 ---
 
 Raised too low fee-payer `maxTotalFee` policy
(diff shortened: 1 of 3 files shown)


brendanjryan (approved):
can we just add a comment on these magic numbers ?

PR #331, opened 2026-04-10
feat: add tempo subscriptions

brendanjryan:
## Summary

Added Tempo subscription support with activation, renewal, reusable subscription authorization, adapter management-response handling, compose dispatch hardening, subscription lifecycle race protection, ergonomic Date/period inputs, attribution metadata cleanup, and typed receipt-response sentinel handling.


src/tempo/server/Subscription.ts
```diff
@@ -0,0 +1,325 @@
...
+      if (!identity) return undefined
+
+      const resource = await parameters.getResource({ identity, input, request })
+      const subscriptionIdHint = input.headers.get('Subscription-Id')?.trim() || undefined
+      const matches = (await store.listByIdentityResource(identity.id, resource.id)).filter(
+        (record) => matchesRequest(record, request),
+      )
+      const active = matches.filter((record) => isActive(record))
+
+      const subscription = (() => {
```
brendanjryan: it feels odd we have to have the client pass a subscription id in order to disambiguate, shouldn't the server be able to filter subscriptions to check if one is valid for the given route / user?

src/tempo/subscription/Store.ts
```diff
@@ -0,0 +1,43 @@
+import * as Store from '../../Store.js'
+import type { SubscriptionRecord } from './Types.js'
+
+const recordPrefix = 'tempo:subscription:record:'
+const resourcePrefix = 'tempo:subscription:resource:'
```
brendanjryan: allow these to be customized

src/tempo/subscription/Store.ts
```diff
@@ -0,0 +1,43 @@
...
+    },
+    async listByIdentityResource(identityId, resourceId) {
+      const ids = ((await store.get(resourceKey(identityId, resourceId))) ?? []) as string[]
+      const records = await Promise.all(
+        ids.map(async (subscriptionId: string) => store.get(recordKey(subscriptionId))),
+      )
+      return records.filter((record: unknown): record is SubscriptionRecord => Boolean(record))
+    },
+    async put(record) {
+      await store.put(recordKey(record.subscriptionId), record)
```
brendanjryan, line 273: add comments about what we are doing here

src/tempo/server/Subscription.ts
```diff
@@ -0,0 +1,325 @@
...
+  }
+}
+
+function formatBillingInterval(periodSeconds: string) {
+  switch (Number(periodSeconds)) {
+    case 60:
+      return 'minute'
+    case 3_600:
+      return 'hour'
+    case 86_400:
```
brendanjryan: make these constants

src/tempo/server/Subscription.ts
```diff
@@ -0,0 +1,325 @@
...
+        .formatToParts(0)
+        .find((part) => part.type === 'currency')?.value ?? metadata.currency
+
+    return `${symbol}${formatUnits(BigInt(request.amount), metadata.decimals)}`
+  } catch {
+    return `$${request.amount}`
+  }
+}
+
+function formatBillingInterval(periodSeconds: string) {
```
brendanjryan: is there no way to manually bill users e.g. in a cron? we should add this

jithinraj:
Great addition, @brendanjryan.

The main thing I’d pin down early is how a later charge proves linkage to the original authorization window.

A minimal stable tuple for that would already help a lot: authorization ref, cumulative spent, remaining window, and a verifier-facing record ref.

brendanjryan:
@jithinraj not sure I follow -- the auth lives on the subscription itself. what is the verifier in this context?

jithinraj:
Got it, thanks @brendanjryan.

By “verifier-facing” I meant downstream audit / reconciliation, not a separate verifier path inside this PR.

My earlier framing was broader than the current scope. The useful boundary seems simpler: subscription reuse is determined by the server’s lookup key plus `stableBinding()`.

Might be worth a short note around that boundary, mainly to distinguish when `subscriptionId` matters from when reuse is handled by the normal request path.

src/tempo/client/Subscription.ts
```diff
@@ -0,0 +1,155 @@
...
+        expiry: toSubscriptionExpirySeconds(toSubscriptionExpiryDate(request.subscriptionExpires)),
+        keyType: accessKey.keyType,
+        limits: [
+          {
+            token: request.currency as Address,
+            limit: BigInt(request.amount),
+            period: toSubscriptionPeriodSeconds(request),
+          },
+        ],
+      },
```
jxom, line 130: We require consumers to also specify `scopes` post-T3 (only relevant here, not for `local`).

examples/subscription/src/server.ts
```diff
@@ -0,0 +1,175 @@
...
+          subscriptionExpires: request.subscriptionExpires,
+          subscriptionId: id,
+          timestamp: new Date().toISOString(),
+        } satisfies Subscription.SubscriptionRecord
+
+        return {
+          receipt: Subscription.createSubscriptionReceipt(record),
+          subscription: record,
+        }
+      },
```
jxom: I thought most of this could be handled internally

PR #333, opened 2026-04-10
fix: accept delegated tempo proof signers

brendanjryan:
## Summary
- accept zero-dollar Tempo proof credentials signed by an authorized access key when `credential.source` identifies the root wallet DID
- expose public `tempo.Proof.proofSource()` and `tempo.Proof.parseProofSource()` helpers so servers can parse `credential.source` without reimplementing the DID format
- add deterministic regression coverage for delegated proof signers in the Tempo server tests

## Manual Verification
- Confirm the local wallet is ready with `tempo wallet -t whoami`.
- Use a wallet where the root `wallet` address differs from the active `key.address` so the request is signed by a delegated key.
- Start a local endpoint protected by a zero-amount `tempo.charge(...)` challenge.
- Call it with `tempo request -s http://127.0.0.1:<port>`.
- Confirm the request succeeds when `credential.source` identifies the wallet DID even though the proof signature was produced by the delegated key.

### DIFF
--- a/.changeset/public-proof-source-helper.md
+++ b/.changeset/public-proof-source-helper.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Expose public Tempo proof-source helpers so servers can parse `credential.source` without reimplementing the `did:pkh:eip155` format.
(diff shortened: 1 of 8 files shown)


brendanjryan:
dup of #332

PR #341, opened 2026-04-15
fix: respect html text overrides on payment pages

### LINKED ISSUE #335: Tempo HTML payment script should respect text.pay config
**Description**

The Tempo payment method's built-in HTML script hardcodes the pay button as `"Continue with {Tempo SVG logo}"` and ignores the `text.pay` config that the server correctly passes through `Html.init`.

**Current behavior**

```ts
// src/tempo/server/internal/html/main.ts:76
button.innerHTML = 'Continue with <svg aria-label="Tempo" ...>...</svg>'
```

Setting `text.pay` has no effect on the button:

```ts
tempo.charge({
  html: {
    text: { pay: "Buy Now" }
  }
})
// Button still shows "Continue with {Tempo logo}"
```

**Expected behavior**

The button text should update when `text.pay` is set, while preserving the default "Continue with {Tempo logo}" when no override is provided.

**What works**

The data pipeline is correct. Verified each step:

1. `Charge.ts` passes `text` to the html options
2. `Mppx.ts:938` serializes `text` (including `pay`) into the page's JSON data
3. `Html.init` exposes it as `c.text.pay` on the context
4. `main.ts` ignores `c.text.pay` entirely

Other text fields work correctly:
- `text.title` > `<title>` tag
- `text.paymentRequired` > badge label in header
- `text.expires` > expiry prefix

**Design question**

The tricky part is the default behavior. Right now the default button has the Tempo SVG logo baked in. A few options:

1. Always show `{text.pay} {logo}` so custom text still gets the logo ("Buy Now {logo}")
2. Show logo only for the default, plain text for custom. But there's no clean way to distinguish "user set pay to 'Pay'" from "default 'Pay'" on the client side without adding a flag
3. Pass a config flag like `config.customPayText` from the server so the client knows whether to use the branded layout or plain text

Would love to hear the thoughts on this. Happy to open a PR once there's alignment on the approach.

**Context**

Found while building the [MPP Playground](https://mpp-playground.vercel.app/html) HTML builder, which lets developers customize the payment page visually. All other text overrides work, only the button text doesn't update.

brendanjryan:
## Summary
- respect custom `text.pay` on the Tempo HTML payment page while preserving the branded default button label
- make the shared HTML `<title>` follow a custom `paymentRequired` label when `title` is omitted
- add render-level and browser-level regressions for both behaviors

Closes #335


src/tempo/server/internal/html/main.ts
```diff
@@ -73,8 +73,9 @@ const provider = Provider.create({
 })
 
 const button = document.createElement('button')
+const buttonLabel = c.text.pay === 'Pay' ? 'Continue with' : c.text.pay
```
brendanjryan, line 76: pulling this out of SVG

.changeset/fix-html-text-overrides.md
```diff
@@ -0,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Fix Tempo HTML pay button text overrides and make the HTML page title follow a custom `paymentRequired` label when `title` is omitted.
```
tmm:
```suggestion
Fixed Tempo HTML pay button text overrides and make the HTML page title follow a custom `paymentRequired` label when `title` is omitted.
```

tmm (approved):

PR #343, opened 2026-04-16
fix: allow zero close for untouched channels

bpierre:
## Summary
- allow `close(0)` when a session channel is still untouched (`spent = 0`, `settled = 0`)
- keep the existing strict `> settled` rule once settlement has advanced
- add a regression test covering zero-close on an untouched channel

## Context
An untouched channel should be able to cooperatively close at `0` instead of falling back to a forced close grace period.

## Testing
- add a regression test in `src/tempo/server/Session.test.ts`


brendanjryan (approved):

brendanjryan:
@bpierre -- this looks good, but need to fix ci

bpierre:
@brendanjryan merged main in :+1:

PR #370, opened 2026-04-20
feat(cli): add account private-key command

brendanjryan:
## Summary
- add `mppx account private-key` for exporting local keychain-backed account keys
- return a clear error for Tempo wallet pseudo-accounts, which are not stored in the local mppx keychain
- add CLI coverage for success and missing-account cases

### DIFF
--- a/.changeset/account-private-key-command.md
+++ b/.changeset/account-private-key-command.md
@@ -2,4 +2,4 @@
 'mppx': patch
 ---
 
-Add an `mppx account private-key` command for exporting the private key of local keychain-backed accounts.
+Added `mppx account private-key` command for exporting the private key of local keychain-backed accounts.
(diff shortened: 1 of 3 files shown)


.changeset/account-private-key-command.md
```diff
@@ -0,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Add an `mppx account private-key` command for exporting the private key of local keychain-backed accounts.
```
tmm:
```suggestion
Added `account private-key` command for exporting the private key of local keychain-backed accounts.
```

src/cli/cli.ts
```diff
@@ -713,6 +713,38 @@ const account = Cli.create('account', {
       }
     },
   })
+  .command('private-key', {
```
tmm: Maybe we call it `export` instead? Allows for other formats in the future.

brendanjryan: 👍

PR #377, opened 2026-04-20
feat: add `acceptPaymentPolicy` to control Accept-Payment header injection

jxom:
## Summary

Adds `acceptPaymentPolicy` option to control when the `Accept-Payment` header is injected on outgoing requests, mitigating CORS preflight failures on non-payment-aware servers.

## Breaking Change

In **browser environments**, polyfilled fetch (`Fetch.polyfill` / `Mppx.create` with `polyfill: true`) now defaults to `'same-origin'` instead of injecting `Accept-Payment` on every request. This prevents CORS preflights from being triggered on cross-origin servers that don't know about `Accept-Payment`.

**Non-browser environments are unaffected** — they continue to default to `'always'` since CORS is not a concern.

## API

```ts
// Mppx.create (browser: defaults to 'same-origin', non-browser: defaults to 'always')
Mppx.create({
  methods: [tempo({ account })],
  acceptPaymentPolicy: 'same-origin',
})
```

### Supported values

- `'always'` — inject on every request (default for `Fetch.from`, and polyfill in non-browser environments)
- `'same-origin'` — inject only for same-origin requests (default for polyfill in browsers)
- `'never'` — never auto-inject
- `{ origins: ['https://pay.example.com', '*.example.com'] }` — inject only for listed origins (supports wildcard subdomains)

### DIFF
--- a/.changeset/accept-payment-policy.md
+++ b/.changeset/accept-payment-policy.md
@@ -4,5 +4,7 @@
 
 Added `acceptPaymentPolicy` option to control when the `Accept-Payment` header is injected on outgoing requests, mitigating CORS preflight failures on non-payment-aware servers.
 
-- `Fetch.polyfill` and `Mppx.create` (with `polyfill: true`) default to `'same-origin'`, preventing cross-origin CORS issues.
-- `Mppx.create` with `polyfill: false` defaults to `'always'`.
+- In browsers, `Fetch.polyfill` and `Mppx.create` (with `polyfill: true`) default to `'same-origin'`, preventing cross-origin CORS issues.
+- Non-browser environments and `Mppx.create` with `polyfill: false` default to `'always'`.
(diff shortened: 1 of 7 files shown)


.changeset/accept-payment-policy.md
```diff
@@ -0,0 +1,8 @@
+---
+'mppx': patch
+---
+
+Added `acceptPaymentPolicy` option to control when the `Accept-Payment` header is injected on outgoing requests, mitigating CORS preflight failures on non-payment-aware servers.
```
brendanjryan, line 5: do we strictly need two patches here?

brendanjryan:
👍 I like this design, and implementation LGTM!

.changeset/accept-payment-policy.md
jxom: One is a breaking change. Can probably consolidate though - will amend!

PR #393, opened 2026-04-25
fix(tempo): bump ox to 0.14.18 so bundled UI preserves limits[].period

### LINKED ISSUE #392: Bundled Tempo payment UI strips limits[].period from on-chain KeyAuthorization
Access keys authorized with a per-period spending limit (e.g. `10 pathUSD per day`) can't silent-sign on mppx-served payment pages. The wallet approval dialog opens instead, because the chain rejects the tx with a `KeyAuthorization` signer-recovery error.

## What's happening

The `KeyAuthorization` is signed correctly with `period` at authorization time. But when mppx's bundled Tempo button builds the `eth_estimateGas` / `eth_sendTransaction` call, `period` gets dropped from `limits[]`:

```json
// What was signed (in IDB):
"limits": [{ "token": "0x20c0…", "limit": "0x989680", "period": 86400 }]

// What goes on-chain:
"limits": [{ "token": "0x20c0…", "limit": "0x989680" }]
```

The chain recomputes the typed-data digest from the period-less payload, gets a different hash than what was signed, and can't recover the signer:

```
Revm error: failed to recover signer from KeyAuthorization signature
```

## Where it comes from

`pnpm-workspace.yaml` overrides pin `ox: '0.14.7'`, which predates `period` in `KeyAuthorization` (the field was added around `0.14.12`). The bundled serializer therefore doesn't know about the field and silently drops it when encoding.

The override is also out of sync with the main `package.json`, which declares `"ox": "0.14.15"`. The override wins, so `0.14.7` is what actually runs everywhere.

## Repro

1. Authorize a Tempo access key with `limits[0].period: 86400` (per day)
2. Open a payment link served via `tempo.charge`
3. Click Pay → approval dialog opens with `eth_sendTransaction` instead of signing silently

Limits without `period` work fine because the signed digest and the on-chain payload have the same shape.

## Fix

Aligning both `ox` declarations on `0.14.18`:

- `pnpm-workspace.yaml`: override `0.14.7` → `0.14.18`
- `package.json`: `"ox": "0.14.15"` → `"ox": "0.14.18"`

Both sources agree, the override stays exact-pin, and the bundled serializer handles `period`. Tested locally against mpp-playground.

starc007:
Closes #392.

The Tempo payment bundle's `KeyAuthorization` serializer was running `ox 0.14.7` (forced by the workspace override), which predates the `period` field. Per-period access keys signed off-chain with a digest that included `period` were re-encoded without it before chain submission, so signer recovery failed:

    Revm error: failed to recover signer from KeyAuthorization signature

Silent signing then fell back to the wallet approval dialog.

## What changed

Aligned both `ox` declarations on `0.14.18`:

- `pnpm-workspace.yaml`: override `0.14.7` → `0.14.18`
- `package.json`: `0.14.15` → `0.14.18`

## Verified locally

- `pnpm check:types` passes for the main package
- `pnpm check:types:html` passes for the bundle subpackages
- Manual test against mpp-playground: silent signing on payment-link pages now works for keys with a per-period limit

## Notes

- `accounts` in `src/tempo/server/internal/html/package.json` is unchanged. Bumping it is unrelated to this bug and would carry the dialog-host migration as separate scope.


brendanjryan:
LGTM -- thank you!

PR #399, opened 2026-04-28
fix: `compose()` JSDoc example

evchip:
The example called `mppx.tempo.charge({...})` but the impl destructures `[methodOrKey, options]` tuples. Updated the docs to match impl

### DIFF
--- a/.changeset/fix-compose-jsdoc-example.md
+++ b/.changeset/fix-compose-jsdoc-example.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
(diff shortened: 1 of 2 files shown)


brendanjryan:
LGTM -- I don't think we need a patch release just for a docs change

evchip:
> LGTM -- I don't think we need a patch release just for a docs change

removed. thanks @brendanjryan

brendanjryan:
LGTM -- ty!

brendanjryan (approved):

PR #401, opened 2026-04-28
feat(tempo): add stateful virtual address charge routing

brendanjryan:
## Context

Adds Tempo virtual-address recipient routing for charge flows with a stateful resolver API. When `virtualAddress: { masterId, resolveTag }` is configured on the server charge method, `externalId` resolves the recipient by default, `virtualAddressId` overrides it, and an explicit `recipient` still wins unless it is combined with `virtualAddressId`, which throws. 

`resolveTag` is the single source of truth for mapping IDs to 48-bit user tags, which avoids the collision risk of stateless 6-byte hashing while keeping the storage strategy in user space.

```mermaid
flowchart TD
    input[Charge request plus virtualAddress config] --> choose{Which recipient source applies?}
    choose -->|virtualAddressId| routeId[Use virtualAddressId]
    choose -->|externalId and no explicit recipient| external[Use externalId]
    choose -->|explicit recipient or no ids| fallback[Use existing recipient path]
    routeId --> resolve[Call virtualAddress.resolveTag]
    external --> resolve
    resolve --> validate[Validate 48-bit unsigned tag]
    validate --> build[Build VirtualAddress from masterId and tag]
    build --> recipient[Use derived virtual recipient]
    fallback --> recipient
```


brendanjryan:
going to close this as not merging right now -- will re-evaluate later

PR #408, opened 2026-05-03
fix: short-circuit express management responses

### LINKED ISSUE #404: Middleware does not handle session management responses (topUp/close)
### Check existing issues

- [x] I checked there isn't [already an issue](https://github.com/wevm/mppx/issues) for the bug I encountered.

### Describe the bug

Summary
The Express middleware (mppx/express) does not correctly handle session-mode management operations (topUp, close). When respond() returns a short-circuit Response (e.g. 204 No Content), the middleware ignores it, calls next(), and lets the user's route handler run — resulting in incorrect responses for management actions.
Reproduction
tsimport express from 'express'
import { Mppx, tempo } from 'mppx/express'

const mppx = Mppx.create({
  methods: [tempo.session({ /* config */ })],
})

const app = express()

app.post('/api/chat', mppx.session({ amount: '0.01' }), (req, res) => {
  // This handler runs for ALL verified requests,
  // including topUp and close management operations.
  res.json({ data: 'content' })
})
When a client sends a topUp or close credential to /api/chat:
Expected: Client receives 204 No Content + Payment-Receipt header (as defined by respond() in tempo/server/Session.ts).
Actual: Client receives 200 OK + { "data": "content" } because next() is called and the route handler responds with business data.
Root Cause
In src/middlewares/express.ts, the payment() function only branches on result.status === 402:
tsreturn async (req, res, next) => {
  const request = new Request(/* ... */)
  const result = await intent(options)(request)

  // Branch 1: 402 → return challenge
  if (result.status === 402) {
    const challenge = result.challenge as Response
    res.status(challenge.status)
    for (const [key, value] of challenge.headers) res.setHeader(key, value)
    res.send(await challenge.text())
    return
  }

  // Branch 2: everything else → patch res.json, call next()
  const originalJson = res.json.bind(res)
  res.json = (body: any) => {
    const wrapped = result.withReceipt(Response.json(body))
    res.setHeader('Payment-Receipt', wrapped.headers.get('Payment-Receipt')!)
    return originalJson(body)  // ← always sends the user handler's body
  }

  next()  // ← topUp/close requests fall through to user handler
}
The problem is that result.withReceipt() internally checks whether respond() returned a management response. If it did, withReceipt() returns the management response (204) instead of the caller's response. But the express middleware ignores the return value of withReceipt() — it only extracts the Payment-Receipt header, then unconditionally calls originalJson(body) with the user handler's payload.
In contrast, the native Fetch API handler pattern (used by all official session examples) works correctly:
ts// Native pattern — withReceipt short-circuit works as intended
return result.withReceipt(Response.json({ data: 'content' }))
// → For topUp/close: returns 204 (respond()'s response), ignores Response.json(...)
// → For voucher/open: returns 200 + JSON body + Payment-Receipt header
Suggested Fix
The middleware should check whether withReceipt() produced a management response, and if so, write it directly to res instead of calling next(). One approach:
tsreturn async (req, res, next) => {
  const request = new Request(/* ... */)
  const result = await intent(options)(request)

  if (result.status === 402) {
    const challenge = result.challenge as Response
    res.status(challenge.status)
    for (const [key, value] of challenge.headers) res.setHeader(key, value)
    res.send(await challenge.text())
    return
  }

  // Probe withReceipt with a sentinel to detect management responses
  const sentinel = new Response(null, { status: 200 })
  const probed = result.withReceipt(sentinel)

  if (probed !== sentinel && probed.status === 204) {
    // Management response (topUp/close) — short-circuit
    res.status(probed.status)
    for (const [key, value] of probed.headers) res.setHeader(key, value)
    res.end()
    return
  }

  // Content response — patch res.json and fall through
  const originalJson = res.json.bind(res)
  res.json = (body: any) => {
    const wrapped = result.withReceipt(Response.json(body))
    res.setHeader('Payment-Receipt', wrapped.headers.get('Payment-Receipt')!)
    return originalJson(body)
  }

  next()
}
Alternatively, withReceipt could expose metadata (e.g. result.isManagementResponse) to make the check explicit without needing a sentinel probe.
Environment

mppx: main branch (latest as of 2026-04-30)
Affected file: src/middlewares/express.ts
Session method: tempo.session()

Notes

All three official session examples (multi-fetch, sse, ws) use the native Request → Response handler pattern, which handles this correctly. Only the Express middleware path is affected.
The charge intent is not affected since it has no management operations.

### Link to Minimal Reproducible Example

_No response_

### Steps To Reproduce

_No response_

### Package Version

0.6.7

### Anything else?

_No response_

brendanjryan:
## Summary

Fix Express middleware to short-circuit session management responses returned by `withReceipt()` and send their status, headers, and body directly before route handlers run.

Add a regression test covering a management `204` response with `Payment-Receipt`.

Add a patch changeset for the Express middleware fix.

Fixes #404.

### DIFF
--- a/.changeset/fresh-blue-keys.md
+++ b/.changeset/fresh-blue-keys.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Fixed Express middleware to short-circuit Tempo session management responses before route handlers run.
--- a/src/middlewares/express.test.ts
+++ b/src/middlewares/express.test.ts
@@ -253,6 +253,33 @@ describe('session', () => {
 })
 
 describe('payment', () => {
+  test('short-circuits management responses', async () => {
+    let handlerRan = false
+    const managementResponse = new Response(null, {
+      status: 204,
+      headers: { 'Payment-Receipt': 'management-receipt' },
+    })
+    const intent = () => async () => ({
+      status: 200 as const,
+      withReceipt: () => managementResponse,
+    })
+
+    const app = express()
+    app.get('/', payment(intent as any, {} as any), (_req, res) => {
+      handlerRan = true
+      res.json({ data: 'content' })
+    })
+
+    const server = await createServer(app)
+    const response = await globalThis.fetch(server.url)
+    expect(response.status).toBe(204)
+    expect(response.headers.get('Payment-Receipt')).toBe('management-receipt')
+    expect(await response.text()).toBe('')
+    expect(handlerRan).toBe(false)
+
+    server.close()
+  })
+
   test('returns 402 when no credential', async () => {
     const { mppx } = createCoreChargeHarness(false)
 
--- a/src/middlewares/express.ts
+++ b/src/middlewares/express.ts
@@ -76,6 +76,30 @@ export function payment<const intent extends Mppx_internal.AnyMethodFn>(
       return
     }
 
+    const managementResponse = (() => {
+      try {
+        return (result.withReceipt as () => Response)()
+      } catch (error) {
+        if (
+          error instanceof Error &&
+          error.message === 'withReceipt() requires a response argument'
+        )
+          return null
+        throw error
+      }
+    })()
+
+    if (managementResponse) {
+      res.status(managementResponse.status)
+      for (const [key, value] of managementResponse.headers) res.setHeader(key, value)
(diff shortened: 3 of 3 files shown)


returnGuo:
@brendanjryan maybe we should check elysia, hono and next.js also.

PR #415, opened 2026-05-06
feat: add EVM charge support for ERC-20 payment auth

douglance:
## Summary
- Adds an `evm/charge` auth method that lets servers gate endpoints with ERC-20 token payments (EIP-3009 / Permit2)
- Wires it into the CLI as a new `evm` plugin and into the client/server method registries
- Ships an `examples/evm-charge` app and anvil-backed integration tests under `test/evm`

## Test plan
- [ ] `pnpm test` (unit + Receipt tests)
- [ ] `pnpm test` against anvil for `src/evm/Charge.anvil.test.ts`
- [ ] Run `examples/evm-charge` end-to-end with a funded private key + RPC


douglance:
unintentional

PR #425, opened 2026-05-09
feat: add `--network` option

o-az:
Adds `--network mainnet|testnet` and checks the challenge chain against the connected RPC before signing.

This came up while dogfooding `mppx` with Hermes Agent. The agent hit a testnet MPP ping challenge but was configured with the mainnet RPC, and later hit a mainnet `fal` challenge while trying the testnet RPC. In both cases the CLI could still sign/broadcast on the wrong chain, which turned into confusing errors like token initialization failures, payment rejected, or verification failed.

Now the CLI fails before signing if the challenge says one chain and the RPC is connected to another.

### DIFF
--- a/.changeset/cli-network-chain.md
+++ b/.changeset/cli-network-chain.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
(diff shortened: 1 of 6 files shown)


brendanjryan (approved):
LGTM

PR #426, opened 2026-05-09
feat: add tempo autoswap cli options

o-az:
Wires autoswap through the CLI.

This came up while dogfooding `mppx` with Hermes Agent against `fal`. The `fal` endpoint requested `USDC.e` on Tempo mainnet, but the funded account had `PathUSD`. Autoswap already existed in the client, but there was no CLI path to turn it on or choose the source token.

Adds `--auto-swap`, `--pay-with`, `--slippage`, plus the matching `-M` opts. Also rejects unknown method opts now, so things like `-M feeToken=...` don't look like they worked when they were actually ignored.

### DIFF
--- a/.changeset/cli-autoswap-options.md
+++ b/.changeset/cli-autoswap-options.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Added Tempo CLI auto-swap and payment source token options.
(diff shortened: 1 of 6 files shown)


brendanjryan (approved):
LGTM

PR #427, opened 2026-05-09
feat: add services registry commands

o-az:
Adds `mppx services` commands for the registry.

This came up while dogfooding mppx with Hermes Agent. The registry data existed at `mpp.dev/api/services`, but the CLI had no way to list services or inspect endpoints, so the agent had to curl the full JSON and filter it manually.

This adds commands to list services, inspect a service, and show its endpoints from the CLI.


brendanjryan (approved):
LGTM

PR #437, opened 2026-05-12
ci: shard runtime tests

brendanjryan:
## Summary

Sharded runtime tests across node and CLI jobs while preserving a single aggregate required check.


brendanjryan:
Duplicate of #436, which is merged.

PR #456, opened 2026-05-18
feat(tempo): add `tempo.subscription.renew`; bump default fee payer gas

jxom:
- Adds `mppx.tempo.subscription.renew({ subscriptionId })` for background subscription renewals on a configured subscription handler. The handler keeps the registered store, client, fee payer, confirmation behavior, and hooks attached through framework wrappers.

  ```diff
  - await tempo.renewSubscription({
  -   ...billing,
  -   store,
  -   subscriptionId,
  - })
  + await mppx.tempo.subscription.renew({
  +   subscriptionId,
  + })
  ```

- Raises the default subscription fee payer policy to cover access-key provisioning, which can cost around 4M gas. Callers can still override `feePayerPolicy` per subscription renewal.

- Removes the extra `transfer` access-key authorization scope from subscriptions. Subscription access-key auth now only requests the `transferWithMemo` scope used for subscription settlement.

### DIFF
--- a/src/tempo/server/Subscription.ts
+++ b/src/tempo/server/Subscription.ts
@@ -41,6 +41,7 @@ import type {
 } from '../subscription/Types.js'
 
 type SubscriptionRequest = ReturnType<typeof Methods.subscription.schema.request.parse>
+type RenewalHandler = NonNullable<subscription.Parameters['renew']>
 
 /**
  * Creates a Tempo subscription method for recurring TIP-20 token payments.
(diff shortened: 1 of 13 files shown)


src/tempo/server/Subscription.ts
```diff
@@ -290,6 +309,19 @@ export function subscription<const parameters extends subscription.Parameters>(
       return activation.result.receipt
     },
   })
+  return method
+}
+
+// Access-key provisioning can cost around 4M gas.
+const defaultFeePayerPolicy = {
```
brendanjryan, line 316: 👍

brendanjryan (approved):

PR #469, opened 2026-05-19
fix(tempo): evict stale session channels

### LINKED ISSUE #467: MPP sessions reuse stale channels after channel-not-found
## Summary
MPP session auto-management can keep using a channel after the service rejects it with `session/channel-not-found`.

Moved from tempoxyz/accounts#520 because the stale channel cache and retry behavior live in MPPX client session management, not the Accounts SDK.

## Expected
- If a session open fails verification, the client should not keep that channel as reusable.
- If a voucher request returns `410 Channel Not Found`, the client should evict the cached channel and reopen on the next request.
- Apps should not need a hard refresh to clear the in-memory session channel map.

## Actual
- The session client can cache a channel after creating an open payload.
- A later request can skip open and ask for a voucher against that channel.
- The service can return `410` with `https://paymentauth.org/problems/session/channel-not-found`, leaving the app stuck until refresh.

## Product impact
The user signs a voucher, pays attention to wallet prompts, and then nothing happens. This makes paid agent flows look unreliable and makes it hard for apps to recover without exposing low-level session state.

## Accounts context
Accounts wires MPPX through `src/core/Provider.ts` via `mppx_tempo({ getClient })`. The recovery behavior should be fixed in the MPPX session client; Accounts can add integration coverage once the upstream client evicts stale channels.

deodad:
## Summary
Evicts cached Tempo session channels after the server reports `session/channel-not-found`.

## Motivation
Fixes #467. A session manager could keep signing vouchers against a channel after the service returned `410 Channel Not Found`, leaving apps stuck until refresh.

## Changes
- Added an internal `createSessionController()` in `src/tempo/client/Session.ts` so session cache state can sync from server responses.
- Clear cached channel maps when a `410` `https://paymentauth.org/problems/session/channel-not-found` response is observed.
- Updated `src/tempo/client/SessionManager.ts` to delegate response sync to the session controller across fetch, open, voucher, and close paths.
- Added a regression test in `src/tempo/server/Session.test.ts` covering eviction and reopen on the next request.

## Testing
- `./node_modules/.bin/tsgo -b src/tsconfig.json --pretty false`
- `./node_modules/.bin/vp test src/tempo/client/SessionManager.test.ts`
- `./node_modules/.bin/vp test src/tempo/server/Session.test.ts -t "evicts cached channel"`
- `./node_modules/.bin/vp fmt --check src/tempo/client/Session.ts src/tempo/client/SessionManager.ts src/tempo/server/Session.test.ts .changeset/fix-stale-session-channel.md`
- `./node_modules/.bin/vp lint src/tempo/client/Session.ts src/tempo/client/SessionManager.ts src/tempo/server/Session.test.ts`

### DIFF
--- a/src/tempo/client/Session.ts
+++ b/src/tempo/client/Session.ts
@@ -71,6 +71,15 @@ export type SessionContext = z.infer<typeof sessionContextSchema>
  * ```
  */
 export function session(parameters: session.Parameters = {}) {
+  return createSessionController(parameters).method
+}
+
+/**
+ * Creates a session method with its internal response handling controls.
+ *
+ * @internal
+ */
+export function createSessionController(parameters: session.Parameters = {}) {
   const { decimals = defaults.decimals } = parameters
 
   const getClient = Client.getResolver({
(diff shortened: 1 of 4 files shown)


src/tempo/client/Session.ts
```diff
@@ -94,6 +103,41 @@ export function session(parameters: session.Parameters = {}) {
     parameters.onChannelUpdate?.(entry)
   }
 
+  function evictChannel(channelId: string): boolean {
+    const key =
+      channelIdToKey.get(channelId) ??
+      Array.from(channelIdToKey).find(
```
brendanjryan, line 109: can you add a comment here to what we are doing and why? this method is a little gnarly

src/tempo/client/Session.ts
```diff
@@ -341,24 +385,46 @@ export function session(parameters: session.Parameters = {}) {
...
-    context: sessionContextSchema,
+  return {
+    method: Method.toClient(Methods.session, {
+      context: sessionContextSchema,
 
-    async createCredential({ challenge, context }) {
-      const chainId = challenge.request.methodDetails?.chainId ?? 0
-      const client = await getClient({ chainId })
-      const account = getAccount(client, context)
+      async createCredential({ challenge, context }) {
```
brendanjryan, line 392: is this intentional?

src/tempo/client/Session.ts
```diff
@@ -341,24 +385,46 @@ export function session(parameters: session.Parameters = {}) {
...
+    syncFromResponse,
+  }
+}
+
+type SessionSyncResult = {
+  channelEvicted: true
+  channelId: string
+}
+
+const ChannelNotFoundProblemType = 'https://paymentauth.org/problems/session/channel-not-found'
```
brendanjryan, line 416: this can likely live alongside other errors / problem types -- we should centralize this since its part of the core control flow

PR #473, opened 2026-05-21
feat: add stripe connect settlement options

brendanjryan:
## Summary

Added server-side Stripe Connect settlement options for Stripe charge PaymentIntent creation.

### DIFF
--- a/src/stripe/server/Charge.test.ts
+++ b/src/stripe/server/Charge.test.ts
@@ -143,7 +143,7 @@ describe('stripe.charge with client', () => {
             return {
               applicationFeeAmount: 12,
               onBehalfOf: 'acct_merchant',
-              stripeAccount: 'acct_platform',
+              stripeAccount: 'acct_connected',
               transferData: { amount: 88, destination: 'acct_destination' },
               transferGroup: 'order_123',
             }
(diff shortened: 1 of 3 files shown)


src/stripe/server/Charge.ts
```diff
@@ -187,6 +200,40 @@ export declare namespace charge {
     parameters,
     Extract<keyof parameters, keyof Defaults>
   > & { decimals: number }
+
+  type ConnectSettlement = {
+    /** Connected account used as the Stripe account context for the request. */
```
logananderson-stripe, line 210: Would it be possible to add the stripe docs link here?

src/stripe/server/Charge.test.ts
```diff
@@ -126,6 +131,176 @@ describe('stripe.charge with client', () => {
...
+    const server = Mppx.create({
+      methods: [
+        stripe.charge({
+          client,
+          connect({ request }) {
+            expect(request.amount).toBe('100')
+            return {
+              applicationFeeAmount: 12,
+              onBehalfOf: 'acct_merchant',
+              stripeAccount: 'acct_platform',
```
logananderson-stripe: nit: stripeAccount is normally the connected account not the platform

src/stripe/server/Charge.ts
```diff
@@ -288,3 +368,48 @@ function buildAnalytics(parameters: { credential: Credential.Credential }): Reco
...
+
+  if (settlement.transferData !== undefined) {
+    validateRequiredAccountId(settlement.transferData.destination, 'transferData.destination')
+    validateAmount(settlement.transferData.amount, paymentAmount, 'transferData.amount')
+  }
+
+  return settlement
+}
+
+function validateAccountId(value: string | undefined, name: string) {
```
logananderson-stripe, line 400: Its awesome to have these basic validations but I am assuming we return an error if the payment intent fails to create?  For example, if the platform doesn't have access to the merchant or something like that.

logananderson-stripe (approved):
Looks great! Left a couple of comments but nothing blocking

PR #482, opened 2026-05-26
Add `secretKey` to session SSE `server.ts` example

ramosdiegocom:
Prevents error thrown when attempting to run server.ts without setting MPP_SECRET_KEY as an environment variable.

```
Error: Missing secret key. Set the MPP_SECRET_KEY environment variable or pass `secretKey` to Mppx.create().
```

The [session WebSocket example](https://github.com/wevm/mppx/blob/main/examples/session/ws/src/server.ts#L51-L54) uses the same pattern. I copied it from there. It makes it easier to get started testing.

### DIFF
--- a/examples/session/sse/src/server.ts
+++ b/examples/session/sse/src/server.ts
@@ -90,6 +90,11 @@ const store = Store.memory()
 
 // Mppx Server Instance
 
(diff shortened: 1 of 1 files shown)


brendanjryan:
ty!

PR #496, opened 2026-06-01
fix: require captured request for session verification

brendanjryan:
## Summary

Requires captured request context for standalone Tempo session voucher and open verification.


brendanjryan:
abandoning this for now since we are launching sessions v2 soon

PR #532, opened 2026-06-10
feat: wallet-bound proof signatures

stevencartavia:
Part of OSS-307

Bind Tempo zero-amount proof signatures to the payer wallet (add `account` to the EIP-712 `Proof` type, domain v3), closing access-key cross-account replay. Expose the proof contract via `tempo.Proof` and add conformance vectors.

### DIFF
--- a/src/tempo/client/Charge.ts
+++ b/src/tempo/client/Charge.ts
@@ -77,10 +77,14 @@ export function charge(parameters: charge.Parameters = {}) {
       if (BigInt(amount) === 0n) {
         const signature = await signTypedData(client, {
           account,
-          domain: Proof.domain(chainId),
-          types: Proof.types,
-          primaryType: 'Proof',
-          message: Proof.message(challenge.id, challenge.realm),
+          // `account` here is the signing account; the proof's bound payer is
+          // `account.address` (echoed in the credential `source` below).
+          ...Proof.typedData({
+            account: account.address,
+            chainId,
+            challengeId: challenge.id,
+            realm: challenge.realm,
+          }),
         })
         return Credential.serialize({
           challenge,
--- a/src/tempo/server/Charge.test.ts
+++ b/src/tempo/server/Charge.test.ts
@@ -2719,6 +2719,91 @@ describe('tempo', () => {
       },
     )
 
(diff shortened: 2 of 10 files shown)


src/tempo/client/Charge.ts
```diff
@@ -77,10 +77,14 @@ export function charge(parameters: charge.Parameters = {}) {
       if (BigInt(amount) === 0n) {
         const signature = await signTypedData(client, {
           account,
-          domain: Proof.domain(chainId),
-          types: Proof.types,
-          primaryType: 'Proof',
-          message: Proof.message(challenge.id, challenge.realm),
+          // `account` here is the signing account; the proof's bound payer is
+          // `account.address` (echoed in the credential `source` below).
+          ...Proof.typedData({
```
figtracer, line 82: can we add a zero-amount client test with an access key? this path relies on `account.address` being the root payer, but the current access-key client tests only cover transaction credentials.

src/tempo/server/Charge.test.ts
```diff
@@ -2583,7 +2591,11 @@ describe('tempo', () => {
             domain: Proof.domain(chain.id),
             types: Proof.types,
             primaryType: 'Proof',
-            message: Proof.message(challenge.id, challenge.realm),
+            message: Proof.message({
```
figtracer, line 2594: can we add a replay test where one access key is authorized for accounts a and b, signs a proof for a, then claims source b? that would cover the wallet-binding regression end to end.

src/tempo/client/Charge.ts
stevencartavia: [2c53aca](https://github.com/wevm/mppx/pull/532/commits/2c53acaf878732fd17d205dacc36eb3b4caf2faa)

src/tempo/server/Charge.test.ts
stevencartavia: [2c53aca](https://github.com/wevm/mppx/pull/532/commits/2c53acaf878732fd17d205dacc36eb3b4caf2faa)

mablr (approved):

brendanjryan (approved):

PR #534, opened 2026-06-11
feat: simulate co-signed transaction before sponsored broadcast

stevencartavia:
The Tempo fee-payer pre-broadcast simulation now simulates the co-signed transaction the sponsor actually broadcasts (concrete fee payer + chosen fee token) instead of the pre-cosign `0x78` envelope, and fails closed (no broadcast) when the simulation reverts. Adds revert and ABI conformance fixtures.

### DIFF
--- a/.changeset/sponsor-presimulate-cosigned.md
+++ b/.changeset/sponsor-presimulate-cosigned.md
@@ -2,4 +2,4 @@
 'mppx': patch
 ---
 
-The Tempo fee-payer (sponsor) pre-broadcast simulation now simulates the co-signed transaction the sponsor actually broadcasts — with the concrete fee payer and chosen fee token — instead of the pre-cosign `0x78` envelope. This catches reverts in the exact transaction the sponsor pays gas for, and fails closed (no broadcast) when the simulation reverts.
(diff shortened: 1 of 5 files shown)


brendanjryan:
Nice! could you rebase this? I made a similar fix earlier this morning and this conflicts

stevencartavia:
@brendanjryan updated!

src/tempo/server/Charge.ts
```diff
@@ -428,11 +444,11 @@ export function charge<const parameters extends charge.Parameters>(
               return serializedTransaction
```
mablr: The hosted fee-payer path still simulates the old `FeePayer.simulationTransaction(..., { feePayer: true })` shape because `simulationRequest` is only overwritten in the local `feePayerAccount` branch. For `feePayerUrl`, the final broadcasted transaction can include the hosted sponsor’s chosen `feeToken`/fee-payer signature, but the pre-broadcast `eth_call` remains calls-only, so this keeps the same sponsor-risk gap for hosted fee payers.

.changeset/sponsor-presimulate-cosigned.md
```diff
@@ -0,0 +1,5 @@
+---
```
mablr, line 1: is this necessary ?

src/tempo/server/Charge.ts
stevencartavia: [8c3863a](https://github.com/wevm/mppx/pull/534/commits/8c3863a07735cbbf6b080897088e31f7f47c770b)

.changeset/sponsor-presimulate-cosigned.md
stevencartavia: it requires a changeset

mablr (approved):

brendanjryan (approved):

PR #526, opened 2026-06-09
feat!: unify mcp client payment wrapping into a single McpClient.wrap

parvahuja:
## Summary

- **Unify MCP client payment wrapping into a single `McpClient.wrap`** (per review: one obvious API). `wrap` now adds payment handling **in place** — the client is mutated and the same reference is returned, so surfaces that keep using the original client become payment-aware (e.g. when another SDK owns the reference, like Cloudflare Agents). Re-wrapping replaces the payment configuration instead of stacking.
- `callTool` keeps the exact MCP SDK `(params, resultSchema?, options?)` signature; `context` and a per-call `onPaymentRequired` (pass `null` to bypass the configured hook) ride in the third argument and are stripped before forwarding.
- Payment challenges are handled whether they arrive as `-32042` errors or as successful tool results carrying `org.paymentauth/payment-required` metadata; payment-required data is schema-validated. Config gains `orderChallenges` and `paymentPreferences`; `methods` accepts tuples.
- Move MCP entrypoints to `mppx/mcp/client` and `mppx/mcp/server`; `mppx/mcp-sdk/*` remain as aliases.

**Breaking** (minor bump, per maintainer sign-off): the previous copy-semantics `wrap` and its `(params, options)` signature are replaced by the in-place SDK-shape API above, and the approval-hook-as-first-argument overload from #522 moves to the options argument.

## Validation

- `pnpm check:ci`, `pnpm check:types`, `pnpm build`; full vitest suite green in CI including localnet runtime shards.
- Export map verified: `mppx/mcp-sdk/client` and `mppx/mcp/client` resolve to the identical module.
- Exercised end-to-end by a deployed Cloudflare Worker paying per tool call on Tempo testnet.

Companion PRs: tempoxyz/mpp#677 (docs), wevm/mppx#540 + tempoxyz/accounts#645 (`wallet_authorizeChallenge`).

PR #540, opened 2026-06-11
feat: add wallet_authorizeChallenge support

parvahuja:
## Summary

- Add `wallet_authorizeChallenge` wallet support: when a Tempo payment method resolves a JSON-RPC account, mppx first asks the wallet to satisfy the Challenge via the `wallet_authorizeChallenge` RPC and falls back to local signing when unsupported.
- Capability-gated: probes `wallet_getCapabilities` for `mpp.status === 'supported'` on the challenge chain. The probe is opportunistic — any probe failure keeps the legacy local-signing path; only the `wallet_authorizeChallenge` call itself falls back strictly on `4200`/`-32601` and rethrows other errors.
- Wire format (v0): `wallet_authorizeChallenge({ challenges: [serialized] })` → `{ authorization }`, validated to answer the exact challenge sent.
- Hooks: `charge` (after client/account resolution and the `expectedRecipients` check, which now also applies to wallet-handled payments) and v2 `session` (automatic mode only; manual actions, caller-supplied channels, and locally opened channels stay on the local path). Sessions probe once per instance, not per voucher.
- Local accounts and `src/tempo/legacy` are untouched; zero new RPCs for non-JSON-RPC accounts.

Supersedes #527, rebuilt on the TIP-1034 session architecture — server session snapshots + pluggable storage replace the old `session` continuation parameter, so the RPC is challenges-in, credential-out.

Wallet-side counterpart (serves the RPC + advertises the capability): tempoxyz/accounts PR (linked below). Independent of #526 — either can land first; clients fall back gracefully until both sides exist.

## Validation

- `pnpm check:types`, `pnpm check:ci`; 53 tests across the helper (table-driven fallback matrix), charge, and session hooks.
- Cross-checked literal-for-literal against the accounts handler; end-to-end exercised against a fake EIP-1193 wallet implementing the accounts dispatch (supported → wallet credential; unsupported → local signing).

### DIFF
--- a/src/tempo/internal/wallet.ts
+++ b/src/tempo/internal/wallet.ts
@@ -1,0 +1,105 @@
+import { numberToHex, type Address, type Client, type Hex } from 'viem'
+
+import * as Challenge from '../../Challenge.js'
+import * as Credential from '../../Credential.js'
+
+/** `wallet_getCapabilities` result, keyed by hex chain ID. */
+type Capabilities = Record<
+  string,
+  { mpp?: { status?: string | undefined } | undefined } | undefined
+>
+
+/**
+ * Asks the wallet behind a JSON-RPC account to satisfy an MPP challenge via
+ * the `wallet_authorizeChallenge` RPC method, gated on a `wallet_getCapabilities` probe.
+ *
+ * Returns `undefined` when the capability probe fails for any reason, when
+ * the wallet does not advertise MPP support, or when the wallet rejects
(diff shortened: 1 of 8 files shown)


src/tempo/internal/wallet.ts
jxom, line 1: maybe `wallet.ts` a better module name?

jxom (approved):

jxom (approved):

PR #566, opened 2026-06-18
feat: verify Tempo subscription transfers credit recipient

stevencartavia:
Under Tempo T6 (TIP-1028), a blocked transfer still succeeds but funds go to `ReceivePolicyGuard`. Confirmed subscription activation/renewal now checks the `TransferWithMemo` log actually credits the recipient instead of trusting tx success.

### DIFF
--- a/.changeset/verify-subscription-renewal-transfer.md
+++ b/.changeset/verify-subscription-renewal-transfer.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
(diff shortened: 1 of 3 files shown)


figtracer (approved):
LGTM from my review pass. No actionable comments.

mablr (commented):
Sg, pending conformance check fix.

brendanjryan (approved):

PR #567, opened 2026-06-18
feat: settle MCP payments in the payment-aware fetch

parvahuja:
## Summary

- Compose `Transport.http()` from protocol handlers for MPP, x402, and MCP-over-HTTP challenges behind the payment-aware fetch.
- Collect all payment offers from HTTP responses and route retry credentials through the protocol that produced the selected challenge.
- Detect MCP JSON-RPC `-32042` responses at the fetch boundary, including the first SSE event, and retry with credentials in MCP `_meta` while preserving Request-carried bodies.
- Harden MCP-over-HTTP message parsing for invalid JSON-RPC response shapes and case-insensitive content types.

## Validation

- `pnpm check:ci`
- `VITE_TEMPO_NETWORK=none pnpm check:types`
- `VITE_TEMPO_NETWORK=none pnpm vp test --project node src/client/internal/protocols/Mcp.test.ts src/client/Transport.test.ts`
- `VITE_TEMPO_NETWORK=none pnpm vp test --project node src/client/internal/Fetch.test.ts -t "combined MPP|MCP-over-HTTP|Request input|native HTTP 402|JSON-RPC payment|PAYMENT-REQUIRED"`

### DIFF
--- a/src/client/Transport.ts
+++ b/src/client/Transport.ts
@@ -1,18 +1,11 @@
 import * as Challenge from '../Challenge.js'
-import * as Constants from '../Constants.js'
 import * as Credential from '../Credential.js'
 import * as Mcp from '../Mcp.js'
-import * as x402_Header from '../x402/Header.js'
-import * as x402_ChallengeBrand from '../x402/internal/ChallengeBrand.js'
-import * as x402_Types from '../x402/Types.js'
-
-const paymentRequiredStatus = 402
-const credentialHeaders = [
-  Constants.Headers.authorization,
-  x402_Types.paymentRequiredHeader,
-  x402_Types.paymentResponseHeader,
-  x402_Types.paymentSignatureHeader,
-]
+import { mcp as mcpProtocol } from './internal/protocols/Mcp.js'
+import { mpp as mppProtocol } from './internal/protocols/Mpp.js'
+import type { Protocol } from './internal/protocols/Protocol.js'
+import { paymentRequiredStatus } from './internal/protocols/Shared.js'
+import { x402 as x402Protocol } from './internal/protocols/X402.js'
 
 /**
  * Client-side transport adapter.
@@ -23,12 +16,21 @@ const credentialHeaders = [
 export type Transport<in out request = unknown, in out response = unknown> = {
   /** Transport name for identification. */
   name: string
-  /** Checks if a response indicates payment is required. */
-  isPaymentRequired: (response: response) => boolean
+  /**
+   * Checks if a response indicates payment is required. May inspect the request (to gate
+   * body reads) and be async (to read a response body).
+   */
+  isPaymentRequired: (response: response, request?: request) => boolean | Promise<boolean>
   /** Extracts all challenges from a payment-required response, when the transport supports multiple offers. */
-  getChallenges?: (response: response) => Challenge.Challenge[]
+  getChallenges?: (
+    response: response,
+    request?: request,
+  ) => Challenge.Challenge[] | Promise<Challenge.Challenge[]>
   /** Extracts the challenge from a payment-required response. */
-  getChallenge: (response: response) => Challenge.Challenge
+  getChallenge: (
+    response: response,
+    request?: request,
+  ) => Challenge.Challenge | Promise<Challenge.Challenge>
   /** Attaches a credential to a request. */
   setCredential: (
     request: request,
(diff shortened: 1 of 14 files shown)


src/client/internal/McpChallengeBrand.ts
```diff
@@ -0,0 +1,14 @@
+const mcpChallenge = Symbol.for('mppx.mcp.challenge')
```
parvahuja: intentionally modeling this as a "challenge brand" similar to how we treat x402 within normal http transport

src/client/Transport.ts
```diff
@@ -75,23 +85,38 @@ export function from<request, response>(
...
- * - Sends credentials via `Authorization` or `PAYMENT-SIGNATURE`
+ * - **402** — Payment-auth challenges from `WWW-Authenticate`, x402 challenges from
+ *   `PAYMENT-REQUIRED`; credential sent via `Authorization` / `PAYMENT-SIGNATURE`.
+ * - **MCP-over-HTTP** — remote MCP rides Streamable HTTP, so its challenge is a JSON-RPC
+ *   `-32042` error in the body of a normal 200 (often `text/event-stream`); the credential
+ *   rides in `_meta`. The `-32042`/`_meta` protocol is delegated to the {@link mcp} codec;
+ *   this transport only adapts the HTTP body. The body is read only when the *request* is a
+ *   JSON-RPC message, so non-MCP traffic (e.g. a streaming LLM call) is never read.
  */
 export function http() {
```
parvahuja: I think longer term we need to break out / refactor the different transports, since they are getting overloaded and now are misnomers. But since fetch and others default to the http transport, the most minimal change rn is to just overload this transport even more.

brendanjryan: What do you envision here? 

Technically http is the right transport, but today we do mix multiple protocols (rest, mcp, etc...) inside the code which is a bit gnarly

IMO it may make sense to have a pluggable "adapter" so that you can add new protocols to a transport easily?

parvahuja: Yeah exactly. Like `http` is the right transport, but then we also have an `mcp` transport (which is a protocol, not a transport). Agree with the pluggable adapter.

parvahuja: This would get rid of the challenge branding stuff too. We would just attempt to dispatch the request/response to all the adapters greedily choose the first one that fits. Should we bundle those changes in this PR?

src/client/internal/adapters/Mcp.ts
```diff
@@ -0,0 +1,79 @@
+import * as Mcp from '../../../Mcp.js'
+import type { Adapter, Transport } from '../../Transport.js'
+import { paymentRequiredStatus } from './Shared.js'
```
brendanjryan: instead of adapters can we call these protocols?

brendanjryan: e.g. src/client/internal/protocols 

we should also comment what these are -- it's not super apparent unless you are familiar with convos on this pr

src/client/internal/adapters/Mpp.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * MPP adapter — the native HTTP payment scheme. Challenges arrive via `WWW-Authenticate` on a
+ * 402; the credential is sent via `Authorization`. As the default flavor it claims nothing
+ * positively (`owns` is false), so the transport routes any challenge no other adapter owns
+ * to it.
+ */
+export function adapter(): Adapter {
+  return {
+    name: 'mpp',
```
brendanjryan: given that we ar pulling from shard anyway, can we make these names constants?

src/client/internal/protocols/Mcp.ts
```diff
@@ -0,0 +1,112 @@
...
+function parseMessage(value: unknown): Mcp.Response | undefined {
+  return value &&
+    typeof value === 'object' &&
+    'jsonrpc' in value &&
+    ('result' in value || 'error' in value)
+    ? (value as Mcp.Response)
+    : undefined
+}
+
+async function parseSseJsonRpcResponse(response: Response): Promise<Mcp.Response | undefined> {
```
brendanjryan, line 80: does the MCP SDK not have utils for doing this? this feels pretty fragile

src/client/internal/protocols/Mcp.ts
```diff
@@ -0,0 +1,112 @@
...
+    setCredential(request, credential) {
+      const message = JSON.parse(request.body as string) as Mcp.Request
+      const parsed = Credential.deserialize(credential)
+      return {
+        ...request,
+        body: JSON.stringify({
+          ...message,
+          params: {
+            ...message.params,
+            ['_meta']: { ...message.params?.['_meta'], [Mcp.credentialMetaKey]: parsed },
```
brendanjryan, line 156: 👍

brendanjryan (approved):
One comment, but otherwise LGTM

src/client/internal/protocols/Mcp.ts
parvahuja: MCP SDK does not afaik... We could add a new dependency like `eventsource-parser` if we wanted get rid of some of this code wdyt?

src/client/internal/protocols/Mcp.ts
```diff
@@ -0,0 +1,159 @@
...
+  if (id === undefined) return undefined
+  const headers = new Headers(request?.headers)
+  if (headers.has('mcp-method')) return id
+  const accept = headers.get('accept')?.toLowerCase() ?? ''
+  return accept.includes('application/json') && accept.includes('text/event-stream')
+    ? id
+    : undefined
+}
+
+function parseMessage(value: unknown): Mcp.Response | undefined {
```
brendanjryan, line 35: can we add tests for this?

brendanjryan: should be able to trivially add table driven tests over a variety of messages shapes

src/client/internal/adapters/Mcp.ts
parvahuja: addressed

src/client/internal/adapters/Mpp.ts
parvahuja: addressed

src/client/internal/protocols/Mcp.ts
parvahuja: added the parser - addressed

src/client/internal/protocols/Mcp.ts
parvahuja: addressed

brendanjryan (approved):

PR #568, opened 2026-06-18
Remove wallet_authorizeChallenge support

parvahuja:
## Summary

- Removed the `wallet_authorizeChallenge` wallet probe path from Tempo charge and session clients.
- Deleted the internal wallet authorization helper, contract tests, and unreleased changeset.
- Kept charge and session challenge signing on the local planning/signing path.

## Validation

- `pnpm test src/tempo/client/Charge.test.ts src/tempo/session/client/Session.test.ts`
- `pnpm check:types`
- `pnpm check:ci`
- `pnpm audit --ignore-registry-errors`
- `git diff --check origin/main...HEAD`


brendanjryan:
why remove the changeset? was this never released?

parvahuja:
> why remove the changeset? was this never released?

just ran a git revert which got rid of it... added a new changeset instead.

PR #571, opened 2026-06-19
feat(session): add client channel store

parvahuja:
## Summary

- Added a client `channelStore` for reusable TIP-1034 session channels.
- Resumes stored channels by payment scope and evicts stale entries before opening fresh.
- Removed the client-side `authorizedSigner` override; voucher authority now comes from the selected account.

## Validation

- `VITE_TEMPO_NETWORK=none pnpm exec vp test --run src/tempo/session/client/CredentialState.test.ts src/tempo/session/client/Session.test.ts src/tempo/session/client/ChannelOps.test.ts src/tempo/session/client/SessionManager.test.ts`
- `pnpm check:types`
- `pnpm check`

### DIFF
--- a/src/client/Mppx.test-d.ts
+++ b/src/client/Mppx.test-d.ts
@@ -1,4 +1,4 @@
-import type { Account } from 'viem'
+import type { Account, Address } from 'viem'
 import { describe, expectTypeOf, test } from 'vp/test'
 
 import * as Challenge from '../Challenge.js'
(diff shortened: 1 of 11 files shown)


src/tempo/session/client/CredentialState.ts
```diff
@@ -533,9 +512,22 @@ export function resolveRecoveredCumulative(
   return settled + requestAmount
 }
 
+/** Returns whether `account` can satisfy the descriptor's voucher authority. */
+export function canSignDescriptor(
+  account: ViemAccount,
+  descriptor: Channel.ChannelDescriptor,
+  authorizedSigner?: Address | undefined,
```
deodad: not directly related but noticed this felt dense and I think `authorizedSigner` is the smell here and as part of the session

mppx is built on passing in an account that is then used, so passing in an address override that can't even sign but we have to check against everywhere seems pointless

think it could likely be dropped and would simplify things a bit. if we really wanted to override the voucher signer it'd make more sense for something like `voucherSigner: Account` but probably fine to assume account == voucher signer for now

deodad (commented):
lgtm though think authorizedSigner could probably go might be worth seeing if brendan has an opinion

src/tempo/session/client/CredentialState.ts
parvahuja: talked with brendan in office and we agreed to just remove the `authorizedSigner` optionality on the client side for now... if this becomes a feature in the future, agree that `voucherAccount` is the better abstraction... updating

src/client/Mppx.test-d.ts
```diff
@@ -24,6 +24,23 @@ describe('Mppx', () => {
     expectTypeOf(tempo.session.manager({ account: {} as Account })).toHaveProperty('fetch')
   })
 
+  test('does not expose a client-side authorizedSigner override', () => {
```
deodad: probably not needed

src/tempo/session/client/CredentialState.ts
```diff
@@ -81,21 +49,31 @@ export function readCredentialCumulativeAmount(
...
   payload: SessionCredentialPayload,
-): void {
-  const key = cache.channelIdToKey.get(channelId)
+): Promise<void> {
   const cumulativeAmount = readCredentialCumulativeAmount(payload)
-  if (!key || cumulativeAmount === undefined) return
-  const entry = cache.channels.get(key)
+  if (cumulativeAmount === undefined) return
+  const entry = await sink.store.get(key)
   if (!entry) return
```
deodad, line 72: This updates the stored entry by payment scope, but manual actions can target a different channel in the same scope. That can advance or delete the current reusable channel when the payload `channelId` does not match. I would no-op unless the stored entry is the exact credential channel:

```suggestion
  const cumulativeAmount = readCredentialCumulativeAmount(payload)
  if (cumulativeAmount === undefined) return
  const entry = await sink.store.get(key)
  if (!entry) return
  if (entry.channelId.toLowerCase() !== payload.channelId.toLowerCase()) return
```

parvahuja: yeah nice catch updated and added test

deodad (approved):

PR #600, opened 2026-06-30
feat: add hint field to PaymentError problem details

bensandler-stripe:
Adds an optional `hint` to the problem details response body for `PaymentRequiredError`, `MalformedCredentialError`, and `PaymentMethodUnsupportedError`. We can consider adding hints for other error types in the future. I expect this to significantly help an agent without prior context on MPP too better unblock themself.

Do I need to do anything to make the same change in our other MPP SDKs?

### DIFF
--- a/src/Errors.ts
+++ b/src/Errors.ts
@@ -1,3 +1,13 @@
+/** Default hint messages for payment errors. */
+export const Hints = {
+  paymentRequired:
+    'Use a supported wallet to pay for this resource using one of the supported payment methods returned in the WWW-Authenticate header. See https://mpp.dev/tools/wallet.md',
+  malformedCredential:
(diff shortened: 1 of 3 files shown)


src/Errors.ts
```diff
@@ -1,3 +1,13 @@
+/** Default hint messages for payment errors. */
+export const Hints = {
+  paymentRequired:
+    'Use a supported wallet to pay for this resource using one of the supported payment methods returned in the WWW-Authenticate header. See https://mpp.dev/tools/wallet.md',
+  malformedCredential:
```
brendanjryan, line 5: 👍

brendanjryan (approved):
LGTM -- this is back compat so we can move to other SDKs ad-hoc

PR #603, opened 2026-07-01
fix(tempo): viem token call

tmm:
Fixes breaking change with token `.call` construction introduced by https://github.com/wevm/viem/releases/tag/viem%402.54.0


src/tempo/client/Charge.ts
```diff
@@ -35,6 +36,23 @@ const chargeContextSchema = z.object({
   mode: z.optional(z.enum(Methods.chargeModes)),
 })
 
+// TODO: Remove once the minimum viem version is >=2.54.0, which uses client-first Tempo call builders.
+function getTransferCall(
```
tmm, line 40: We should keep this wrapper function around for a bit even though Viem peer dep is bumped to `>=2.54.0` since this still keeps backwards compat with older Viem versions (+ gives folks a warning that their Viem version is out of peer dep range). In a month or so, should be good to remove.

src/tempo/internal/auto-swap.ts
```diff
@@ -14,6 +15,38 @@ export const defaultCurrencies: readonly Address[] = [
   defaults.tokens.usdc as Address,
 ]
 
+// TODO: Remove once the minimum viem version is >=2.54.0, which uses client-first Tempo call builders.
+function getBalanceCall(
```
tmm, line 19: Same comment as above

src/tempo/internal/auto-swap.ts
```diff
@@ -14,6 +15,38 @@ export const defaultCurrencies: readonly Address[] = [
...
+    (
+      client: Client,
+      parameters: viem_token.getBalance.Args,
+    ): ReturnType<typeof viem_token.getBalance.call>
+  }
+  return call.length >= 2 ? call(client, parameters) : call(parameters)
+}
+
+// TODO: Remove once the minimum viem version is >=2.54.0, which uses client-first Tempo call builders.
+function getApproveCall(
```
tmm, line 35: Same comment as above

brendanjryan (approved):

brendanjryan:
Code LGTM -- types and eslint seem not happy?

PR #612, opened 2026-07-02
feat: preserve method-specific receipt extension fields

IkerAlus:
The core spec's Payment-Receipt section allows method specifications to define additional receipt fields, but `Receipt.Schema` strips unknown fields on parse. Hence servers can't emit them (`Receipt.from`) and no mppx client can read them (`Receipt.fromResponse`, the fetch client, the CLI).

This PR loosens the runtime schema while keeping the exported `Receipt` type unchanged:

- Receipt.Schema is now z.looseObject(shape); unknown fields survive from / serialize / deserialize / fromResponse round-trips. Known-field validation is unchanged.
- Producers pass extension fields through Receipt.from; consumers type them via intersection (Receipt.Receipt & { originTxHash: string }).

**Motivation**: the new `nearintents` method (cross-chain settlement) requires `originTxHash` and `challengeId` in its receipt (a cross-chain payment has two legs, and reference only carries the destination-side tx). Our plan is to ship the method as a standalone package per the custom-payment-methods guide. This is the one core touchpoint a package can't work around, because the stripping happens on the client side.

Also, the repo already assumes extended receipts in places, the CLI renders arbitrary receipt fields (and explorer-links txHash), and tempo session receipts carry channelId / acceptedCumulative / spent. This aligns the core schema with that.

**Backward compatible**: existing methods emit unchanged receipts; older clients strip extras exactly as today. Tests added for preservation, round-trip, and validation-with-extensions; changeset included.

### DIFF
--- a/.changeset/receipt-extension-fields.md
+++ b/.changeset/receipt-extension-fields.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Preserved method-specific extension fields on receipts. `Receipt.from`, `Receipt.deserialize`, and `Receipt.fromResponse` previously stripped fields outside the base set; they now pass unknown fields through, per the core spec's Payment-Receipt provision ("Payment method specifications MAY define additional fields for receipts").
--- a/src/Receipt.test.ts
+++ b/src/Receipt.test.ts
@@ -30,6 +30,42 @@ describe('from', () => {
       }),
     ).toThrow()
   })
+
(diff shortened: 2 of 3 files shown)


brendanjryan (approved):
LGTM -- thank you!

brendanjryan:
@IkerAlus -- can you add a changeset?

IkerAlus:
hey @brendanjryan thanks for the prompt review, just added the change set.

PR #617, opened 2026-07-03
fix(session): pin channel open/top-up read-back to receipt block

grandizzy:
After broadcasting an escrow open/top-up, the server re-reads on-chain channel state at `latest` and validates it against the receipt. Behind a load-balanced RPC, that read can land on a replica that hasn't yet imported the block the transaction was mined in, returning empty/stale state. Verification then fails with a spurious `402 on-chain channel state does not match open receipt` even though the transaction succeeded and the caller was charged.

This pins the read-back to the block from the transaction receipt, so a node either returns authoritative state or errors (and is retried) instead of answering with a lagging view. Transient read failures are retried while the replica catches up.

Tests cover block-pinning (read-back targets the receipt block, not `latest`) and the retry helper (success, retry-until-success, and exhaustion).

### DIFF
--- a/.changeset/session-open-readback-block-pin.md
+++ b/.changeset/session-open-readback-block-pin.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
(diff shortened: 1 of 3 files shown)


figtracer (approved):
makes sense

brendanjryan (approved):
Good find -- thank you!

PR #621, opened 2026-07-06
fix(session): disable replacement receipt detection

parvahuja:
## Summary

- Disabled viem replacement detection while waiting for Tempo session precompile receipts.
- Added coverage for consecutive Tempo transactions that reuse the visible sender nonce.

## Why

viem treats same-sender, same-nonce transactions as replacements by default. Tempo 0x76 transactions can reuse the visible nonce across distinct transactions, so a charge followed by a session open could resolve to the earlier charge receipt and fail to find `ChannelOpened`.

## Validation

- `VITE_TEMPO_NETWORK=none pnpm exec vp test --run src/tempo/session/precompile/Chain.test.ts`
- `pnpm check`
- `pnpm check:ci`
- `pnpm check:types`
- `pnpm build`
- Packed install smoke: charge and session returned 200 through `mppx.fetch` and global `fetch`.

### DIFF
--- a/.changeset/fixed-tempo-session-receipts.md
+++ b/.changeset/fixed-tempo-session-receipts.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Fixed Tempo session receipt waits after consecutive same-nonce transactions.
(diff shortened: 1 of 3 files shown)


parvahuja:
cyclops audit fast

PR #622, opened 2026-07-06
Programmatic validation API with path params, x402 detection, and JSON output

bensandler-stripe:
We want to enable other tools (e.g. the WIP `machine-payments-doctor`) to import validation logic directly from mppx rather than reimplementing it, while keeping the CLI as a thin wrapper. To make this possible I extracted validation logic from the CLI into a reusable `src/validate/core.ts` module, exposed as `mppx/validate` package export. This can now be called programmatically (vs only via CLI).

I upstreamed the following from `machine-payments-doctor`:
- OpenAPI path parameter substitution (uses `example` / `schema.example` / `schema.default` from the spec)
- Detect x402 protocol headers (`PAYMENT-REQUIRED` / `X-Payment-Required`) on 402 responses without MPP `WWW-Authenticate`
- Supports multi-path discovery fallback (`/openapi.json`, `/api/openapi.json`) with explicit `--discovery-path` override

I also add a `--outputJson` / `-j` flag for machine-readable validation output (will be useful for evals...).


src/validation/core.ts
```diff
@@ -0,0 +1,251 @@
+import { validateChallenge, validateErrorHandling } from '../cli/validate/challenge.js'
```
brendanjryan, line 1: instead of /validate could this package be `validation`?

src/validation/core.ts
```diff
@@ -0,0 +1,251 @@
...
+  flags: {
+    sawMppEndpoint: boolean
+    sawNonMppPaymentEndpoint: boolean
+    sawTestnet: boolean
+    sawMainnet: boolean
+    paymentSucceeded: boolean
+  }
+}
+
+export async function validate(options: ValidateOptions): Promise<ValidateResult> {
```
brendanjryan, line 58: can we add jsdoc here?

brendanjryan (changes requested):
two tiny comments!

bensandler-stripe:
done!

brendanjryan:
LGTM!

PR #623, opened 2026-07-06
feat(client): support standard x402 exact payments

parvahuja:
## Summary
- support standard x402 V2 `exact` EVM EIP-3009 client payments without requiring mppx route binding
- skip x402 exact offers this client cannot pay so a later valid EIP-3009 offer can be selected
- preserve mppx-bound nonce derivation when the route-binding extension is present

## Scope
Support remains limited to x402 V2 `exact` EVM EIP-3009. Permit2, `upto`, batch settlement, and non-EVM schemes remain unsupported.

### DIFF
--- a/.changeset/standard-x402-exact-client.md
+++ b/.changeset/standard-x402-exact-client.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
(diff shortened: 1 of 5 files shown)


parvahuja:
@brendanjryan wave

PR #629, opened 2026-07-09
mppx validate: multi-challenge support, per-method validation, streaming output

bensandler-stripe:
A number of improvements to `mppx validate`. Builds on #631.

Bug fixes:
- Instead of picking first method for automatic crypto transfer, iterates over all methods and makes transfers for any supported method (tempo, soon to be base)
- Fixed URL resolution when base URL has a subpath (e.g. `mppx validate https://api.example.com/mpp` now correctly resolves to `/mpp/openapi.json` instead of `/openapi.json`)
- Guarded all BigInt conversions with `isValidIntegerAmount` regex check first (prevents crash on non-integer amounts like "5.000000")
- Multi-challenge parsing: uses `fromResponseList` to parse all payment methods from `WWW-Authenticate` (previously only picked the first one)
- Amount is always required (fail not warn when missing)
- Console.log leak fixed in JSON mode

UX and validation improvements:
- Per-method field validation: validates Stripe (amount, currency, networkId, paymentMethodTypes), EVM (recipient, currency, amount, chainId), and Tempo (recipient/splits, currency, amount) challenge fields
- Common field checks reduced across all challenges (id, realm, expiration, realm-match) into single pass/fail per field
- Token symbol and chain name in payment prompts ("USDC.e on Tempo Mainnet")
- Tempo recipient validation: must have either `recipient` or `splits`
- Unsupported payment methods silently skipped (no noise in summary count)

### Diffs before/after

- API which supports multiple methods: https://www.diffchecker.com/MlCLRk6Q
- API which supports one method: https://www.diffchecker.com/2OxdCZuw
- API where we previously completely failed: https://www.diffchecker.com/uAPHWsyM

### DIFF
--- a/.changeset/validate-error-reporting.md
+++ b/.changeset/validate-error-reporting.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+`mppx validate`: Multi-challenge support (parses all payment methods from WWW-Authenticate), per-method field validation for Stripe/EVM/Tempo, streaming CLI output, fixed URL resolution for subpaths, and improved error reporting for malformed challenges.
--- a/src/cli/validate.test.ts
+++ b/src/cli/validate.test.ts
@@ -159,6 +159,12 @@ describe('validate: discovery', () => {
     expect(output).toContain('Document found and parseable')
   })
 
+  test('discovers openapi.json at root when subpath given', { timeout: 15_000 }, async () => {
+    const server = await mppServer(makeChallenge())
+    const { output } = await serve(['validate', `${server.url}/mpp`])
+    expect(output).toContain('Document found and parseable')
+  })
+
   test('reports invalid JSON', { timeout: 15_000 }, async () => {
     const server = await testServer((req, res) => {
       if (req.url?.includes('openapi.json')) {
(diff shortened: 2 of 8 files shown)


brendanjryan (commented):
@bensandler-stripe -- some CI issues but content-wise this look good!

bensandler-stripe:
ah @brendanjryan sorry if I wasted your time, I was adding a few other fixes to this branch. Will get your review in a few.

bensandler-stripe:
@brendanjryan let's merge https://github.com/wevm/mppx/pull/631 first (this PR is now rebased off that branch) and then I'll have you review this one

.changeset/validate-error-reporting.md
```diff
@@ -0,0 +1,5 @@
+---
```
bensandler-stripe, line 1: diff is a bit easier to review with hiding whitespace:

Image

src/cli/validate.test.ts
```diff
@@ -537,3 +543,137 @@ describe('validate: summary', () => {
...
+    expect(output).toContain('Uses splits (no single recipient)')
+    expect(output).not.toContain('Missing recipient')
+  })
+
+  test('fails on invalid Stripe amount', { timeout: 15_000 }, async () => {
+    const bad = makeStripeChallenge()
+    ;(bad.request as Record<string, unknown>).amount = '1.50'
+    const server = await multiChallengeServer([bad])
+    const { output } = await serve(['validate', server.url])
+    expect(output).toContain('Amount is valid integer string (Got: 1.50)')
```
brendanjryan, line 677: why is this "is a valid integer string?" shouldn't this be invalid?

brendanjryan:
code-wise this LGTM 

> Guarded all BigInt conversions with isValidIntegerAmount regex check first (prevents crash on non-integer amounts like "5.000000")


Why do we need checks like this? Shouldn't all of this be implicitly guarded by the `mppx` typing and data types? 

There is some concern with servers passing invalid json on the wire, but IMO we should just fail are struct marshal time vs. trying to do field-level checks. This is much simpler to maintain

bensandler-stripe:
> Why do we need checks like this? Shouldn't all of this be implicitly guarded by the mppx typing and data types?

@brendanjryan you mean, servers' usage of mppx (or another SDK) should prevent this from even happening?

My thinking is, if we rejected at struct-marshal time, we'd just get "parse failed" with no actionable feedback. There was a real user's server I validated ([this diff](https://www.diffchecker.com/uAPHWsyM) shows before/after) where we previously failed on an opaque `Error: Cannot convert 5.000000 to a BigInt` and now return something much more descriptive.

Is your contention that we can't reasonably do this for every field, so perhaps it's not really worth doing for any? I think amounts specifically are reasonably tricky to get right, so I personally feel good about these validations. But I agree I wouldn't want to do the same for every other field.

brendanjryan:
Yes -- main concern being that this is a little noisy and redundant, but I do see the benefit. 

I'm +1 to this given that all of this is isolated in a designated "validate" package

brendanjryan (approved):
@bensandler-stripe -- some CI issues but content-wise this look good!

PR #631, opened 2026-07-10
Stream validate output as each phase completes

bensandler-stripe:
A recent PR (https://github.com/wevm/mppx/pull/622), which exposes our validation logic to non-CLI callers, accidentally regressed `mppx validate` causing it to batch all results and only print them after the entire run finished. This means the payment prompt (which blocks on user input) appeared before any discovery or challenge output was visible.

Example before:
https://github.com/user-attachments/assets/729ce9c2-61c5-45ef-a1fc-a24ebf5ff08c

This PR adds `validateStream()`, an async generator that yields results per-phase. The CLI now uses `for await` so discovery and challenge checks print immediately, and the payment prompt appears in context. `validate()` remains as a batch wrapper for JSON mode and the programmatic API.

Example after:
https://github.com/user-attachments/assets/28fcd9a4-7d9f-49c2-a84a-f5057df5b98d

### DIFF
--- a/.changeset/validate-stream-refactor.md
+++ b/.changeset/validate-stream-refactor.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+`mppx validate`: Stream results as each phase completes instead of batching until the end. Discovery and challenge checks now appear before the payment prompt.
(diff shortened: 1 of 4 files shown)


brendanjryan:
LGTM!

PR #632, opened 2026-07-12
fix(session): open a new channel when the authorizedSigner changes

### LINKED ISSUE #471: Session auto-management reuses channel after authorizedSigner changes
## Summary

In auto-managed Tempo sessions, `mppx` reuses an existing channel when the client account's `authorizedSigner` changes. The next credential is emitted as a voucher for the old channel instead of opening a new channel.

## Repro

This only prepares/signs credentials. It does not submit a transaction.

```ts
import { Challenge, Credential } from 'mppx'
import { session } from 'mppx/client'
import { createClient, custom } from 'viem'
import { generatePrivateKey, privateKeyToAccount } from 'viem/accounts'
import { Account as TempoAccount } from 'viem/tempo'
import { tempoModerato } from 'viem/tempo/chains'

const payer = privateKeyToAccount(generatePrivateKey())
const signerA = TempoAccount.fromSecp256k1(generatePrivateKey(), { access: payer })
const signerB = TempoAccount.fromSecp256k1(generatePrivateKey(), { access: payer })

const payee = privateKeyToAccount(generatePrivateKey()).address
const currency = '0x20c0000000000000000000000000000000000000'
const escrow = '0xe1c4d3dce17bc111181ddf716f75bae49e61a336'

let account = signerA

function getClient() {
  return createClient({
    account,
    chain: tempoModerato,
    transport: custom({
      async request({ method }) {
        if (method === 'eth_chainId') return `0x${tempoModerato.id.toString(16)}`
        if (method === 'eth_fillTransaction')
          return {
            raw: '0x',
            tx: {
              chainId: `0x${tempoModerato.id.toString(16)}`,
              from: payer.address,
              gas: '0x5208',
              input: '0x',
              maxFeePerGas: '0x1',
              maxPriorityFeePerGas: '0x1',
              nonce: '0x0',
              to: escrow,
              type: '0x2',
              value: '0x0',
            },
          }
        throw new Error(`unexpected rpc method: ${method}`)
      },
    }),
  })
}

function challenge() {
  return Challenge.from({
    id: `challenge-${account.accessKeyAddress}`,
    intent: 'session',
    method: 'tempo',
    realm: 'example.test',
    request: {
      amount: '1000000',
      currency,
      methodDetails: { chainId: tempoModerato.id, escrowContract: escrow },
      recipient: payee,
      suggestedDeposit: '10000000',
      unitType: 'token',
    },
  })
}

async function main() {
  const method = session({ getClient, maxDeposit: '10' })

  const first = Credential.deserialize(
    await method.createCredential({ challenge: challenge(), context: {} }),
  )

  account = signerB

  const second = Credential.deserialize(
    await method.createCredential({ challenge: challenge(), context: {} }),
  )

  console.log({
    first: {
      action: first.payload.action,
      authorizedSigner: first.payload.authorizedSigner,
      channelId: first.payload.channelId,
    },
    second: {
      action: second.payload.action,
      authorizedSigner: second.payload.authorizedSigner,
      channelId: second.payload.channelId,
    },
  })
}

main()
```

## Actual

The second credential reuses the first channel and returns a voucher:

```ts
{
  first: {
    action: 'open',
    authorizedSigner: '0x...',
    channelId: '0x...'
  },
  second: {
    action: 'voucher',
    authorizedSigner: undefined,
    channelId: '0x...' // same as first
  }
}
```

## Expected

When the account's `authorizedSigner` changes, auto-managed sessions should not emit a voucher for a channel opened with the previous signer. It should open a distinct channel for the new signer.

saneGuy:
Fixes #471

## Bug
In auto-managed legacy sessions, `session()` reuses an existing channel after the client account's `authorizedSigner` changes. The next credential is emitted as a voucher for the old channel — signed by the new signer, which the escrow rejects — instead of opening a new channel.

## Root cause
The legacy session client caches channels keyed by `payee:currency:escrow` only, and `ChannelEntry` does not record the voucher authority the channel was opened with. `autoManageCredential` therefore has no way to notice that the resolved signer can no longer produce valid vouchers for the cached channel. (The TIP-1034 precompile client already guards this via `canSignDescriptor` in `planCredential`; this brings the legacy client to parity.)

## Fix
- Track `authorizedSigner` on the legacy `ChannelEntry`: `createOpenPayload` records the authority it opens with, and `tryRecoverChannel` records the on-chain authority (normalizing the zero address to the payer, since that is what the escrow verifies vouchers against in that case).
- In `autoManageCredential`, skip reuse of a cached or recovered entry whose authority no longer matches the resolved voucher signer, falling through to the existing open path.

## Tests
- `reuses the cached channel while the voucher signer is unchanged` (guards the reuse path)
- `opens a new channel instead of reusing when the authorizedSigner changes` (fails before this change with the exact reported symptom: voucher for the old channel)
- `tryRecoverChannel` assertions for zero-address → payer normalization and for preserving a non-zero on-chain `authorizedSigner`
- Full suite green against a Tempo localnet; `check:types` and `check:ci` clean

### DIFF
--- a/.changeset/fix-session-signer-channel-reuse.md
+++ b/.changeset/fix-session-signer-channel-reuse.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Fixed legacy session auto-management reusing a cached channel after the account's `authorizedSigner` changed. The client now opens a new channel instead of emitting a voucher the escrow would reject.
--- a/src/tempo/legacy/client/ChannelOps.test.ts
+++ b/src/tempo/legacy/client/ChannelOps.test.ts
@@ -340,6 +340,27 @@ describe.runIf(isLocalnet)('tryRecoverChannel', () => {
     expect(result!.opened).toBe(true)
     expect(result!.escrowContract).toBe(escrow)
     expect(result!.chainId).toBe(chain.id)
+    // Channel opened with a zero authorizedSigner: the escrow verifies
+    // vouchers against the payer, so the recovered authority is the payer.
+    expect(result!.authorizedSigner.toLowerCase()).toBe(payer.address.toLowerCase())
+  })
+
+  test('recovers the on-chain authorizedSigner when set', async () => {
+    const authorizedSigner = accounts[4].address
+    const { channelId: signerChannelId } = await openChannel({
+      escrow,
+      payer,
+      payee,
+      token: currency,
+      deposit: 10_000_000n,
+      salt: Hex.random(32) as `0x${string}`,
+      authorizedSigner,
+    })
+
+    const result = await tryRecoverChannel(client, escrow, signerChannelId, chain.id)
+
+    expect(result).not.toBeUndefined()
+    expect(result!.authorizedSigner.toLowerCase()).toBe(authorizedSigner.toLowerCase())
   })
 
   test('returns undefined for non-existent channel', async () => {
--- a/src/tempo/legacy/client/ChannelOps.ts
+++ b/src/tempo/legacy/client/ChannelOps.ts
@@ -27,6 +27,8 @@ import { signVoucher } from '../session/Voucher.js'
 
 /** Cached channel metadata used by the legacy auto-driving session client. */
 export type ChannelEntry = {
+  /** Voucher authority the channel was opened with. */
+  authorizedSigner: Address
   /** Legacy contract-backed channel ID. */
   channelId: Hex.Hex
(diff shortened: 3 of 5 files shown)


brendanjryan:
LGTM! We want most improvements to go into the non-legacy session, but given that this is a very safe bugfix we will merge

saneGuy:
Thanks Brendan, appreciate the review and the context. Understood on focusing future improvements on the non-legacy session where possible. I kept this one scoped to the reported legacy-session bug.

PR #643, opened 2026-07-14
Handle paid Tempo session SSE fetches

parvahuja:
## Summary
- Added a private client response hook for methods that need to adapt successful paid fetch responses.
- Used that hook for Tempo session SSE responses so payment control events can post top-up and voucher credentials while callers receive the application stream.
- Added focused coverage for the fetch hook and the Tempo session SSE response path.

## Why
Paid Tempo session SSE streams can request additional vouchers after the initial paid retry has already returned a 200 event stream. `Fetch.from` previously returned that stream directly, so callers using payment-aware fetch could receive payment control events without the client posting the required voucher credentials.

## Validation
- `VITE_TEMPO_NETWORK=none pnpm exec vp test --run src/tempo/session/client/Transports.test.ts`
- `VITE_TEMPO_NETWORK=none pnpm exec vp test --run src/client/internal/Fetch.test.ts -t "lets methods wrap post-payment responses"`
- `pnpm check:types`
- `pnpm check`
- `git diff --check`

### DIFF
--- a/.changeset/payment-aware-sse-session-fetch.md
+++ b/.changeset/payment-aware-sse-session-fetch.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Added paid Tempo session SSE handling to payment-aware fetch responses.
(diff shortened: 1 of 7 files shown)


parvahuja:
Closing in favor of a narrower SessionManager account-resolution change.

PR #649, opened 2026-07-15
validate: improve body handling,  resolution, and x402 compat

bensandler-stripe:
Small collection of fixes for `mppx validate` found while testing against real-world servers (postalform, agentmint, milliondollarbotpage).

- Retry challenge probe on 422 (not just 400). Many frameworks return 422 for missing/invalid body instead of 400.
- Send explicit `--body` on the first request instead of always probing bare first. Avoids wasted round-trip when you know the endpoint needs a body.
- Add `--header` / `-H` flag for custom request headers (e.g. `Idempotency-Key`). Some servers require app-specific headers before issuing a 402.
- Resolve `$ref` in OpenAPI schemas when auto-generating request bodies. Previously any schema using `$ref` (the standard) would fail to generate a body.
- Support `examples` (plural, OpenAPI 3.1 media type map) in addition to `example` (singular).
- Fix x402 validation failing on non-EVM networks (Solana, Tempo). The schema was EVM-only, now uses a tolerant envelope decoder.
- Report x402 findings alongside MPP when both headers are present on the same response.


bensandler-stripe:
r? @brendanjryan 

one more `mppx validate` pr!

brendanjryan (approved):

PR #655, opened 2026-07-15
refactor(session): extract SSE response driver

parvahuja:
## Summary

- extract session SSE payment-event handling from the message projection
- preserve non-payment SSE frames for response-based consumers
- keep `SessionManager.sse()` behavior unchanged

## Motivation

The existing SSE path projected responses directly into message strings, so fetch integrations could not reuse session payment handling while preserving an SSE response. This internal driver separates those concerns without changing the public API.

## Validation

- `VITE_TEMPO_NETWORK=none pnpm exec vp test --run src/tempo/session/client/Transports.test.ts src/tempo/session/client/SessionManager.test.ts`
- `VITE_TEMPO_NETWORK=none pnpm exec vp test --run src/tempo/session`
- `pnpm check:types`
- `pnpm check:ci`

### DIFF
--- a/src/tempo/session/client/Transports.ts
+++ b/src/tempo/session/client/Transports.ts
@@ -777,29 +777,49 @@ export async function openSseSession(
   if (!isEventStream(response)) throw new Error('SSE response is not an event stream.')
   if (!response.body) throw new Error('Response has no body.')
 
-  return iterateSseResponse({
-    challenge,
-    driver,
-    input,
-    onReceipt,
+  return iterateSseMessages({
+    onNeedVoucher: (event) => handleSseNeedVoucher({ challenge, driver, input }, event),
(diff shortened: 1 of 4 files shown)


src/tempo/session/client/Transports.ts
```diff
@@ -777,29 +777,49 @@ export async function openSseSession(
   if (!isEventStream(response)) throw new Error('SSE response is not an event stream.')
   if (!response.body) throw new Error('Response has no body.')
 
-  return iterateSseResponse({
-    challenge,
-    driver,
-    input,
-    onReceipt,
+  return iterateSseMessages({
```
brendanjryan, line 780: 👍

brendanjryan (approved):

PR #664, opened 2026-07-17
fix: require configured Stripe request fields

brendanjryan:
## Motivation

Server method factories accept optional request defaults and return a second-stage handler. Previously, an omitted Stripe `decimals` value was represented as a default in TypeScript even though it remained `undefined` at runtime, so a missing value failed only during Zod validation.

## Summary

- Added `ConfiguredDefaults` to carry only factory values known to be defined.
- Added an explicit resolved-defaults layer for Tempo factories, which always resolve `currency` and `decimals`.
- Preserved caller-provided defaults through `tempo.common()`.
- Added compile-time coverage for configured values, widened optional configuration, Tempo defaults, and legacy session defaults.

```ts
const mppx = Mppx.create({
  methods: [stripe.charge({ client, networkId: "internal", paymentMethodTypes: ["card"] })],
  realm,
  secretKey,
})

mppx.charge({ amount: "1", currency: "usd" })
// TypeScript error: `decimals` is required.

mppx.charge({ amount: "1", currency: "usd", decimals: 2 })
// OK
```

```ts
const mppx = Mppx.create({ methods: [tempo.charge()], realm, secretKey })

mppx.tempo.charge({ amount: "1" })
// OK: Tempo resolves `currency` and `decimals` before returning the handler.
```

## Key design considerations

- A handler treats a field as optional only when its factory configuration is statically known and defined, or when the factory unconditionally resolves it at runtime.
- Stripe has no intrinsic decimal-scale fallback, so callers configure `decimals` either on `stripe.charge()` or on each charge request.
- TypeScript protects typed call sites; existing Zod validation remains the safeguard for JavaScript, `any`, and dynamic values.

### DIFF
--- a/.changeset/bright-items-sleep.md
+++ b/.changeset/bright-items-sleep.md
@@ -2,4 +2,4 @@
 'mppx': patch
 ---
 
-Fixed Stripe charge handlers to require request fields that were not configured as defaults.
+Fixed server method handlers to require request fields that were not configured as defaults, and exposed Tempo's resolved currency and decimals as defaults.
--- a/.changeset/bright-items-sleep.md
+++ b/.changeset/bright-items-sleep.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Fixed Stripe charge handlers to require request fields that were not configured as defaults.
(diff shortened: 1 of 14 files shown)


bensandler-stripe (approved):
Thank you for the quick fix, this is going to help a lot based on my evals. As a relative typescript newb, I'm impressed by what you can accomplish with typescript's type system!

PR #667, opened 2026-07-17
Add hook to tempo sessions, triggered on each on-chain tx

raubrey-stripe:
Adding an onSessionSettled hook to fire on each on-chain settlement/transaction (I'm not opinionated on the naming here, whatever is best!).

This:
- Provide visibility into tx's that top-level existing hooks miss (server-driven settlement)
- Provide visibilitly into tx's that manual instrumentation outside of the SDK cannot provide (namely: clients of the SDK are never notified of transactions triggered as the result of an auto-settlement policy firing an on-chain settlement)
- Is useful when syncing each on-chain txHash to an external location (like Stripe, where you might want to manage/accumulate your on-chain volume)

Some thoughts
- While this sits within the tempo method, it attempts to be generic enough to apply across other methods that support sessions (like Solana) without instrumenting those methods.
- I would have preferred a top level event Mppx.onSessionSettled(...) event like I attempted [here](https://github.com/wevm/mppx/commit/514e507d851a6a7101a6927fea0dc22f42b12952), but the code was pretty gnarly and it doesn't feel like I know enough about the ethos of top level events to make that call.


brendanjryan (approved):
LGTM

raubrey-stripe:
FYI @brendanjryan could I get another quick look? 

- Added changeset
- Fixed a gap: in this event we really need the total `amount` and the `delta` of this settlement event. The naming here.. I'm not sure about

brendanjryan (approved):

brendanjryan:
code LGTM -- some CI issues @raubrey-stripe

raubrey-stripe:
one more poke for re-running workflows @brendanjryan - I had to add a signing key for verified commits/fixed some other papercuts

brendanjryan:
ok merging this and will fix ci in post

PR #668, opened 2026-07-19
feat: support dynamic WebSocket session charges

gakonst:
Adds an optional raw-unit amount to the manual WebSocket session controller's `charge` method. This lets bidirectional services price each application message after inspecting it while retaining the existing fixed challenge tick as the default.

This is needed for persistent paid transports such as OpenAI Responses WebSocket, where every `response.create` frame can have a different request-aware price. Reservations are still committed only when the corresponding application message is emitted.

Validation:
- focused WebSocket metering test (Tempo network disabled)
- `pnpm check:types`

### DIFF
--- a/.changeset/dynamic-websocket-session-charges.md
+++ b/.changeset/dynamic-websocket-session-charges.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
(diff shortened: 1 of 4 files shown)


brendanjryan (approved):
LGTM

PR #673, opened 2026-07-20
fix(subscription): repair missing access-key address indexes

EfeDurmaz16:
## What

- Repair a missing subscription access-key address index whenever an existing primary access-key record is reused.
- Preserve any existing address index.
- Add regression coverage and a patch changeset.

## Why

`getOrCreateAccessKey` returned early for an existing primary record. If the subsequent address-index write failed or the index was lost, retries could never heal it, so `getAccessKeyByAddress` kept returning `null`.

## How

Both existing and newly-created records now pass through one idempotent `ensureAccessKeyAddressIndex` helper. It only writes when the index is missing.

## Verification

- Subscription store tests: 21/21
- `pnpm check:ci`
- `pnpm check:types`

### DIFF
--- a/.changeset/repair-subscription-access-key-index.md
+++ b/.changeset/repair-subscription-access-key-index.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Fixed missing subscription access-key address indexes when reusing existing keys.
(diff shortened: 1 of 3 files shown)


brendanjryan (approved):
LGTM -- thanks!

PR #675, opened 2026-07-20
feat(tempo): rehydrate client sessions from server snapshots

gakonst:
## Summary

This PR adds the client-side half of session rehydration. It lets a client that starts without a hydrated local channel store resume a server-known Tempo session instead of abandoning the channel or restarting its cumulative vouchers from zero.

- hydrate client session state from a server-provided snapshot
- restore the server's highest accepted voucher and current required cumulative boundary
- reconcile the snapshot with the current on-chain channel state
- preserve monotonic cumulative vouchers after restart
- include resumed spend when enforcing the channel's maximum-deposit policy
- reject snapshots for the wrong account/channel or with malformed voucher state

The voucher check confirms that the snapshot's `acceptedCumulative` value corresponds to a voucher signed by the client account. The snapshot's `requiredCumulative` may be higher and is treated as the server's current payment requirement, subject to the configured maximum-deposit policy.

This does **not** establish that the server's pricing is fair, and it does not prevent the server from requesting a higher new payment. Independent pricing policy is outside this PR.

This PR deliberately does not add server-side channel discovery/snapshot emission (#676) or the Node SQLite channel store (#677).

## Why this is needed

The latest client can only resume an off-chain session when its local store already contains the channel state. With no hydrated `channels.db`, it does not know the server's highest unredeemed cumulative voucher. On-chain state alone is insufficient because the server may hold a newer accepted voucher that has not been settled.

The snapshot supplies that missing off-chain baseline so a fresh client can continue the session.

## Exact current-main red proof

Tested against exact `origin/main` commit `bd9b05c` before applying this branch:

- a direct recovered session generated cumulative `15` below the server's accepted `1,000,000`; this branch resumes from `1,000,000`
- the manager recovery path generated `400` below accepted `500`; this branch resumes from `500`
- maximum-deposit accounting used `700` instead of the resumed `1,000`; this branch includes resumed spend
- main cached closed-channel and invalid/wrong-voucher snapshots; this branch rejects those inconsistent recovery inputs

These tests use synthetic snapshots and mocked Tempo chain reads, so the client behavior is independently reviewable before the server starts emitting snapshots.

## Validation

- `pnpm check:ci`: pass (existing underscore warnings only)
- `pnpm check:types`: pass
- focused client rehydration regressions: pass
- prior full runtime matrix for these changes: pass

### DIFF
--- a/src/tempo/session/Snapshot.ts
+++ b/src/tempo/session/Snapshot.ts
@@ -22,6 +22,17 @@ export type SessionSnapshot = {
   escrow: Address
   /** Minimum cumulative authorization needed for the challenged request or stream continuation. */
   requiredCumulative: RawAmountString
+  /** Highest client-signed voucher accepted by the server. */
+  highestVoucher?:
+    | {
+        /** Channel identifier bound into the voucher signature. */
+        channelId: Hex
+        /** Cumulative authorization bound into the voucher signature. */
+        cumulativeAmount: RawAmountString
+        /** Original client signature proving the accepted cumulative authorization. */
+        signature: Hex
+      }
+    | undefined
   /** Amount already settled on-chain. */
   settled: RawAmountString
   /** Amount consumed by delivered content according to server accounting. */
@@ -35,6 +46,9 @@ const addressSchema = z.custom<Address>(
 )
 const hashSchema = z.custom<Hex>(
   (value) => typeof value === 'string' && /^0x[0-9a-fA-F]{64}$/.test(value),
+)
+const hexSchema = z.custom<Hex>(
+  (value) => typeof value === 'string' && /^0x[0-9a-fA-F]+$/.test(value),
 )
 
 const channelDescriptorSchema = z.object({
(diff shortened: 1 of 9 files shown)


src/tempo/session/client/CredentialState.test.ts
```diff
@@ -443,7 +443,7 @@ describe('CredentialPlan', () => {
       })
     })
 
-    test('recovery cumulative ignores server-advertised unused voucher headroom', () => {
+    test('recovery cumulative never decreases below the server-accepted voucher', () => {
```
brendanjryan, line 446: 👍

src/tempo/session/Snapshot.ts
```diff
@@ -55,6 +69,13 @@ const sessionSnapshotSchema = z.object({
   deposit: z.string(),
   descriptor: channelDescriptorSchema,
   escrow: addressSchema,
+  highestVoucher: z.optional(
+    z.object({
+      channelId: hashSchema,
```
brendanjryan, line 74: 👍

brendanjryan (approved):

brendanjryan (approved):

PR #677, opened 2026-07-20
feat(client): add Node SQLite channel store

gakonst:
## Summary

Independent PR; it can merge before or after the session-rehydration PRs.

This PR contains only the Node persistence adapter:

- add `mppx/client/node`
- default to `~/.tempo/wallet/channels.db`
- namespace reusable channels by service origin
- persist channel state across manager/process restarts
- preserve monotonic cumulative and deposit values during concurrent/stale writes
- read existing wallet-cli v2 `channels` rows for migration compatibility
- use WAL, a busy timeout, and explicit `close()` lifecycle

There are no session protocol, bootstrap, signature-validation, or server lookup changes in this PR.

## Exact current-main red proof

Tested against exact latest `origin/main` commit `bd9b05c`:

```text
Cannot find module 'mppx/client/node'
```

The identical feature tests pass here for default path selection, persistence/reopen, monotonic merging, service isolation, and legacy wallet-cli row migration.

## Validation

- `pnpm check:ci`: pass (existing underscore warnings only)
- `pnpm check:types`: pass
- SQLite adapter regressions: pass
- prior full runtime matrix for these changes: pass

### DIFF
--- a/.changeset/node-sqlite-channel-store.md
+++ b/.changeset/node-sqlite-channel-store.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Add a Node SQLite session `ChannelStore` that defaults to Tempo Wallet's channels database and can read existing wallet-cli v2 session rows.
--- a/package.json
+++ b/package.json
@@ -96,6 +96,11 @@
       "types": "./dist/client/index.d.ts",
       "default": "./dist/client/index.js"
     },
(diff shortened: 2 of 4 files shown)


brendanjryan (approved):
I think the schema could be cleaner, but this is isolated and independant so can easily be refactored later if need be

PR #684, opened 2026-07-21
fix: restore main CI checks

brendanjryan:
## Motivation

Restored main checks blocked by an empty changeset description and a newly disclosed dependency advisory.

## Summary

- Added the missing description for the Tempo session design changeset.
- Updated the Hono Node server override to the patched release.

## Key design considerations

- Kept the dependency update scoped to the affected transitive package and regenerated the lockfile.


brendanjryan:
(fixed on main)

PR #693, opened 2026-07-22
feat: auto-swap session deposits

gakonst:
## Summary

- reuse the existing Tempo charge auto-swap resolver for TIP-1034 session deposits
- atomically prepend `approve` and `swapExactAmountOut` before session `open` and `topUp`
- validate the swap prefix server-side and bind its exact output to the channel deposit

## Behavior

When `autoSwap` is enabled, sessions use the same fallback ordering and slippage configuration as charge. If the payer already holds enough channel currency, no swap is added. An open acquires the resolved opening deposit; a top-up acquires only the additional deposit. Vouchers and session recovery do not trigger swaps.

The resulting transaction is either:

```
open / topUp
```

or:

```
approve → swapExactAmountOut → open / topUp
```

## Validation

- client tests cover atomic open and top-up call construction
- server tests accept a canonical exact-output swap and reject output that does not match the deposit
- full CI passes
- verified against the hosted OpenAI MPP service with a sponsored mainnet PathUSD → USDC.e session open, followed by session reuse without another swap

### DIFF
--- a/.changeset/session-autoswap.md
+++ b/.changeset/session-autoswap.md
@@ -1,5 +1,5 @@
 ---
-'mppx': minor
+'mppx': patch
 ---
 
 Added automatic stablecoin swaps for Tempo session channel opens and top-ups.
--- a/.changeset/session-autoswap.md
+++ b/.changeset/session-autoswap.md
@@ -1,0 +1,5 @@
+---
+'mppx': minor
+---
+
+Added automatic stablecoin swaps for Tempo session channel opens and top-ups.
--- a/src/tempo/session/precompile/Chain.ts
+++ b/src/tempo/session/precompile/Chain.ts
@@ -1,5 +1,5 @@
 import type { Account, Address, Client, Hex } from 'viem'
-import { encodeFunctionData, isAddressEqual, parseEventLogs } from 'viem'
+import { decodeFunctionData, encodeFunctionData, isAddressEqual, parseEventLogs } from 'viem'
 import {
   call,
   prepareTransactionRequest,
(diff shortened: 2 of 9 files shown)


.changeset/session-autoswap.md
```diff
@@ -0,0 +1,5 @@
+---
+'mppx': minor
```
parvahuja: this should probably be a patch unless we want to cut to 0.9.0?

parvahuja (approved):

.changeset/session-autoswap.md
brendanjryan: +1 can we bump down to patch?

src/tempo/session/precompile/Chain.ts
```diff
@@ -431,7 +432,88 @@ function parsePrecompileCredentialTransaction(parameters: {
...
     })
-  return { transaction, call: { ...call, data: call.data, to: call.to } }
+  return {
+    transaction,
+    call: { ...call, data: call.data, to: call.to },
+    prefixCalls: calls.slice(0, -1),
+  }
+}
+
+function validateAutoSwapPrefix(parameters: {
```
brendanjryan, line 442: this method is a little gnarly, but not bad if we have good test coverage. 

can we add a table driven test here this should be easy to parameterize

.changeset/session-autoswap.md
gakonst: Addressed in 604395f: the changeset now uses `patch` per the pre-v1 versioning policy.

src/tempo/session/precompile/Chain.ts
gakonst: Addressed in 604395f: added a table-driven suite covering 14 invalid auto-swap prefixes (missing fields, native value, targets, calldata/functions, approval/swap binding, output, and canonical encoding). The focused Chain suite passes 60/60.

alexrisch (approved):

PR #698, opened 2026-07-23
feat(tempo): batch automatic session top-ups

gakonst:
## Summary

- add an optional `topUpAmount` policy to the low-level session client and `sessionManager`
- use the configured refill quantum when a voucher crosses the current deposit
- continue to cap every refill at `maxDeposit` and preserve exact-shortfall behavior by default

## Why

Services that omit `suggestedDeposit` currently trigger an on-chain top-up for every request after a channel is exhausted. A client can now choose a refill quantum (for example, $5) without changing the server protocol or the default behavior.

## Validation

- `VITE_TEMPO_NETWORK=none vp test --run src/tempo/session/client/SessionManager.test.ts src/tempo/session/client/Session.test.ts` (70 passed)
- `vp fmt --check` on the three changed files
- `vp lint` on the three changed files (one pre-existing `__test_setChannel` warning)

The repository install is currently blocked by the existing `viem@2.55.6` workspace resolution from the pkg.pr.new override; the focused suite was run from the already-cached packages.

### DIFF
--- a/.changeset/batched-session-top-ups.md
+++ b/.changeset/batched-session-top-ups.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Added an optional `topUpAmount` policy for batching automatic Tempo session top-ups while preserving exact-shortfall behavior by default.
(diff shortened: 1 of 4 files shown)


brendanjryan (approved):
LGTM

PR #699, opened 2026-07-23
fix(tempo): coordinate parallel sponsor fee exposure

gakonst:
## What changed

- replace sender-level serialization with a sponsor-wide atomic fee budget
- key the budget by the recovered on-chain sponsor address, independent of replay-store prefixes
- persist the final sponsor-signed transaction hash and an owner-fenced lifecycle (`prepared` → `broadcasting` → `pending`)
- keep broadcast/pending exposure reserved until a receipt is observed or the expiring nonce reaches `validBefore`
- reconcile receipts lazily before admitting more work; stale workers cannot release replacement reservations
- bound both aggregate declared fee exposure and reservation count
- wait for capacity until the payment challenge deadline rather than rejecting parallel expiring-nonce submissions

## Why

Tempo expiring nonces are designed for independent transactions to be submitted in parallel. A sender mutex defeats that property, while releasing budget immediately after an optimistic RPC response fails to bound transactions that are still pending or whose response was lost.

This keeps the aggregate sponsor bound without coupling independent senders or same-sender expiring-nonce transactions. The final signed hash and owner token provide the useful lifecycle/reconciliation pieces discussed in #690; this PR intentionally does **not** take #690’s sender lock or add a manual recovery API. It supersedes #690 for this behavior.

Tempo references:
- https://docs.tempo.xyz/guide/payments/send-parallel-transactions
- https://docs.tempo.xyz/protocol/transactions

## Policy

`tempo.charge({ feePayerPolicy })` accepts:

- `maxInFlightTotalFee`: aggregate declared worst-case fee exposure; defaults to `maxTotalFee * 10`
- `maxInFlightReservations`: count bound protecting the budget from zero/low-fee request amplification; defaults to `100`

Per-transaction fee validation remains in the shared fee-payer policy. The aggregate settings are charge-broadcast policy and do not leak into session fee-payer configuration. A single transaction larger than the aggregate budget is rejected immediately instead of waiting until expiry.

Capacity waiters perform atomic no-op reads while full. A challenge expiry stops a waiter from beginning new work, but an already broadcast reservation remains counted through the transaction’s later `validBefore` boundary.

## Validation

- reproduced the previous same-sender rejection on pristine `main`
- verified concurrent same-sender expiring-nonce transactions succeed
- verified different senders sharing one sponsor coordinate against the same atomic budget
- verified capacity waits and proceeds after terminal receipt reconciliation
- verified optimistic sponsored broadcasts remain persisted as `pending`
- verified owner fencing rejects stale transitions/releases
- verified reservation-count limiting and oversized-transaction rejection
- local Tempo node: 180 focused fee-payer/charge/budget tests passed
- `pnpm check:ci`
- `pnpm exec tsgo -b`
- `pnpm build`

Both commits are SSH-signed for GitHub verification.

### DIFF
--- a/src/tempo/server/Charge.test.ts
+++ b/src/tempo/server/Charge.test.ts
@@ -1951,25 +1951,25 @@ describe('tempo', () => {
     })
 
     test('behavior: fee payer allows concurrent expiring-nonce transactions from one sender', async () => {
-      let releaseSimulation!: () => void
-      let resolveSimulationStarted!: () => void
-      const simulationStarted = new Promise<void>((resolve) => {
-        resolveSimulationStarted = resolve
+      let releaseBroadcast!: () => void
+      let resolveBroadcastStarted!: () => void
+      const broadcastStarted = new Promise<void>((resolve) => {
+        resolveBroadcastStarted = resolve
       })
-      const releaseSimulationPromise = new Promise<void>((resolve) => {
-        releaseSimulation = resolve
+      const releaseBroadcastPromise = new Promise<void>((resolve) => {
+        releaseBroadcast = resolve
       })
-      let heldFirstSimulation = false
+      let heldFirstBroadcast = false
 
       const interceptingClient = createClient({
         account: accounts[0],
         chain: client.chain,
         transport: custom({
           async request(args: any) {
-            if (args.method === 'eth_call' && !heldFirstSimulation) {
-              heldFirstSimulation = true
-              resolveSimulationStarted()
-              await releaseSimulationPromise
+            if (args.method === 'eth_sendRawTransactionSync' && !heldFirstBroadcast) {
+              heldFirstBroadcast = true
+              resolveBroadcastStarted()
+              await releaseBroadcastPromise
             }
             return client.transport.request(args)
           },
@@ -2031,13 +2031,13 @@ describe('tempo', () => {
       ])
 
       const first = fetch(httpServer.url, { headers: { Authorization: credential1 } })
-      await simulationStarted
+      await broadcastStarted
       const sponsorBudgetKey =
-        `tenant:mppx:charge:sponsor-budget:${chain.id}:${accounts[0].address.toLowerCase()}` as const
+        `mppx:charge:sponsor-budget:${chain.id}:${accounts[0].address.toLowerCase()}` as const
       expect(await sponsoredStore.get(sponsorBudgetKey)).not.toBeNull()
       expect(
         await sponsoredStore.get(
-          `mppx:charge:sponsor-budget:${chain.id}:${accounts[0].address.toLowerCase()}`,
+          `tenant:mppx:charge:sponsor-budget:${chain.id}:${accounts[0].address.toLowerCase()}`,
         ),
       ).toBeNull()
       const second = fetch(httpServer.url, { headers: { Authorization: credential2 } })
@@ -2046,7 +2046,7 @@ describe('tempo', () => {
         const secondResponse = await second
         expect(secondResponse.status).toBe(200)
       } finally {
-        releaseSimulation()
+        releaseBroadcast()
       }
 
       const firstResponse = await first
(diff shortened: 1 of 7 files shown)


src/tempo/server/Charge.ts
```diff
@@ -642,8 +659,8 @@ export function charge<const parameters extends charge.Parameters>(
             if (releaseReservation) await releaseHashUse(store, hash)
             throw error
           } finally {
-            if (sponsoredSenderReservation)
-              await releaseSponsoredSenderInFlight(store, sponsoredSenderReservation)
+            if (sponsorBudgetReservation)
+              await releaseSponsorBudget(store, sponsorBudgetReservation)
```
parvahuja: with `waitForConfirmation: false`, this releases the budget while the transaction is still pending, so the sponsor could exceed the limit - maybe we should hold it until the transaction confirms or expires, although that may be slightly overkill

parvahuja:
outside of the edge case above, i think 👍

src/tempo/server/Charge.test.ts
```diff
@@ -1939,7 +1950,7 @@ describe('tempo', () => {
       httpServer.close()
     })
 
-    test('behavior: fee payer rejects concurrent in-flight transactions from one sender', async () => {
+    test('behavior: fee payer allows concurrent expiring-nonce transactions from one sender', async () => {
```
brendanjryan, line 1953: 👍

src/tempo/server/Charge.ts
```diff
@@ -1095,12 +1117,29 @@ function getProofStoreKey(challengeId: string): `mppx:charge:${string}` {
...
+type SponsorBudgetState = {
+  reservations: Record<string, { expiresAt: number; fee: string }>
+}
+
+type SponsorBudgetReservation = {
+  chainId: number
+  expiresAt: number
+  fee: bigint
+  hash: `0x${string}`
+  sponsor: string
```
brendanjryan: is this an address as well? I think we could do the same `0x$` typing?

src/tempo/server/Charge.ts
```diff
@@ -1136,23 +1175,85 @@ function parseHashCredentialSource(parameters: {
...
-): Promise<boolean> {
-  return store.update(getSponsoredSenderStoreKey(parameters), (current) => {
-    if (current !== null) return { op: 'noop', result: false }
-    return { op: 'set', value: Date.now(), result: true }
-  })
+  parameters: SponsorBudgetReservation & { maxInFlightTotalFee: bigint },
+): Promise<void> {
+  const key = getSponsorBudgetStoreKey(parameters)
+  let pollIntervalMs = sponsorBudgetPollIntervalMs
+  for (;;) {
```
brendanjryan: 😬 can we add a comment that this won't infinitely spin because of :1196? as long as a reasonably expiresAt is set?

src/tempo/server/Charge.ts
```diff
@@ -1136,23 +1175,85 @@ function parseHashCredentialSource(parameters: {
...
+    const now = Date.now()
+    if (now >= parameters.expiresAt)
+      throw new VerificationFailedError({
```
brendanjryan: would feel better if this had more tests

src/tempo/server/Charge.ts
gakonst: Fixed in a034bbef. Optimistic broadcasts now persist the final sponsor-signed transaction hash and remain budgeted in the `pending` phase until receipt reconciliation or the transaction reaches `validBefore`; ambiguous broadcast failures remain in `broadcasting` under the same rule. The new localnet test asserts the optimistic reservation is still present after the RPC response.

src/tempo/server/Charge.ts
gakonst: Yep. The hosted fee-payer response is now cryptographically recovered to its concrete `Hex` sponsor address, and local sponsorship already supplies the account address. Budget keys are therefore always scoped by `chainId + recovered sponsor address`; the URL/hash fallback was removed.

src/tempo/server/Charge.ts
gakonst: Fixed and documented in the helper. Admission waits are bounded by `waitUntil = min(challenge expiry, validBefore)`, so the loop terminates instead of spinning indefinitely. Once broadcast, the separate persisted `expiresAt = validBefore` keeps exposure counted for the full period in which the transaction can still land.

src/tempo/server/Charge.ts
gakonst: Added focused tests for pending-until-receipt reconciliation, stale-owner fencing, reservation-count limiting, and immediate rejection when one transaction exceeds the aggregate budget. Charge integration tests also cover same-sender parallelism, shared sponsor capacity waiting, expired cleanup, recovered sponsor scoping, and optimistic pending persistence. The focused local Tempo suite passes.

PR #709, opened 2026-07-24
fix(tempo): refresh websocket session challenges

gakonst:
Long-lived Tempo session WebSockets reuse the opening challenge for later vouchers. Once the default five-minute challenge expires, an otherwise healthy socket is rejected with `payment-expired`.

This change:
- re-probes the HTTP payment route when the active WebSocket challenge is expired or within five seconds of expiry
- creates later vouchers against the fresh challenge
- rotates client receipt validation and server close-ready bookkeeping to the fresh challenge ID while keeping the same socket/channel

Regression evidence:
- latest-main negative control: opening succeeds, the stream idles past the challenge lifetime, and the later voucher never resumes
- patched test: the same socket re-probes, sends one voucher, and delivers both chunks
- 183 focused client/server session and WebSocket tests pass
- `pnpm check:ci` and `pnpm check:types` pass

Found by a real 25-minute Nanocodex/OpenAI MPP WebSocket stress run, which ended with a terminal `payment-expired` response after its Code Mode phase.

### DIFF
--- a/.changeset/fresh-websocket-vouchers.md
+++ b/.changeset/fresh-websocket-vouchers.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Refreshed expired session challenges before creating later WebSocket vouchers.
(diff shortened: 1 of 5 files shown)


brendanjryan (approved):
oh good find -- LGTM

PR #712, opened 2026-07-24
fix: bump postcss and tar to patched versions

bensandler-stripe:
CI's dependency audit was failing on two known advisories in transitive deps: postcss and tar. The existing tar override was pinned to 7.5.19, itself a vulnerable version. Bumped overrides to postcss 8.5.18 and tar 7.5.21.

This should help resolve CI failures on [my other PR](https://github.com/wevm/mppx/pull/711) (which I'll land after this).


brendanjryan:
ty!

PR #713, opened 2026-07-24
feat(cli): share durable Tempo session state

gakonst:
## What this does

This PR moves durable TIP-1034 client session storage and administration out of the CLI-private implementation and into the public Node APIs. The MPPx CLI then consumes those same APIs for normal paid requests and `mppx sessions` commands.

### Shared Node session APIs

- expands `mppx/client/node` with the payer-scoped `createSqliteChannelStore` and `createSessionAdministration` APIs
- defaults to `~/.tempo/wallet/channels.db` and remains compatible with existing v2 `channels` rows
- isolates reusable channels by payer, service origin, and payment scope
- adds cross-process SQLite leases so concurrent clients serialize work for the same payer/payment scope
- retains superseded and recovered channels so callers can inspect or explicitly select them

### CLI behavior

- uses the shared SQLite store for local/private-key accounts instead of maintaining a second CLI session registry
- preserves `--session auto`, `--session new`, and explicit `--session <channel-id>` selection
- uses the normal server bootstrap flow for cold-client rehydration
- provides `mppx sessions list`, `view`, `sync`, and `close`
- discovers unknown payer channels from `ChannelOpened` logs and reconciles retained rows with canonical precompile state
- supports cooperative close as well as on-chain request-close/finalize flows
- requires `--yes` for `sessions close --all`

### Durable-state safety

- only deletes local state after a cooperative close returns an authoritative MPP receipt
- only advances request-close state after a successful transaction receipt and on-chain readback
- only deletes a finalized session after successful withdrawal and confirmation that the channel is closed

## Scope boundaries

- this does not parse or manage Tempo Wallet `store.json`
- this does not implement wallet login or access-key lifecycle management
- `tempo:*` accounts remain on the existing Tempo Wallet plugin path; local/private-key accounts use the shared Node SQLite path
- MPPx does not add a second rehydration protocol or session registry: recovery uses normal session bootstrap plus canonical on-chain state

The SQLite-backed Node APIs and persistent session CLI paths require Node.js 22.5 or newer. Other CLI commands retain the package's existing Node.js baseline.

## Dependency follow-up included here

- replaces the expired Viem PR preview with Viem `2.55.8`
- raises the Viem peer floor to `>=2.55.7`, the first published release containing the Tempo relay signature fix
- rejects non-primitive subscription authorization signatures before encoding
- resolves the current brace-expansion audit advisory

## Validation

- `pnpm check:ci`
- `pnpm check:types`
- `pnpm build`
- `pnpm check:types:examples`
- `VITE_TEMPO_NETWORK=none pnpm exec vp test run src/client/node.test.ts --project node`
- `VITE_TEMPO_NETWORK=none pnpm exec vp test run src/cli/sessions/request.test.ts src/cli/mcp.test.ts --project cli`
- focused CLI regressions for Tempo Wallet routing, explicit channel selection, and `sessions close --all --yes`

The latest fix commit is signed and verified.


src/cli/TempoWallet.ts
```diff
@@ -0,0 +1,233 @@
+import { readFile } from 'node:fs/promises'
+import { homedir } from 'node:os'
+import { join } from 'node:path'
+
+import { Json, PublicKey } from 'ox'
+import type { Address, Hex } from 'viem'
+import { Account as TempoAccount, KeyAuthorizationManager } from 'viem/tempo'
+
+type StoredAccount = {
```
brendanjryan: why do we need to include tempo-wallet in the Client? can we just not just these data types from the CLI? 

I would prefer we not couple tempo wallet to mppx

.changeset/calm-sessions-share.md
```diff
@@ -0,0 +1,5 @@
+---
```
brendanjryan, line 1: can we consolidate to a single patch

src/cli/TempoWallet.ts
gakonst: Done — #714 removed the Tempo Wallet/store.json implementation from MPPx. The shared client API now takes an injected payer/account and only owns channels.db session state.

.changeset/calm-sessions-share.md
gakonst: Done — all durable-session changes are consolidated into `calm-sessions-share.md`. The other changeset belongs to the separate Viem dependency fix from #714 that was merged into this branch.

src/tempo/subscription/KeyAuthorization.ts
```diff
@@ -168,8 +168,14 @@ export async function signSubscriptionKeyAuthorization(parameters: {
   const signature = await account.sign({
     hash: KeyAuthorization.getSignPayload(authorization),
   })
+  const signatureEnvelope = SignatureEnvelope.from(signature)
```
brendanjryan, line 171: 👍

src/cli/cli.ts
```diff
@@ -1638,6 +1666,174 @@ const discover = Cli.create('discover', {
...
+      all: z.boolean().optional().describe('Close every session'),
+      cooperative: z.boolean().optional().describe('Ask the service to close cooperatively'),
+      finalize: z.boolean().optional().describe('Withdraw every finalizable session'),
+      header: z.array(z.string()).optional().describe('Add close request header (repeatable)'),
+      onChain: z.boolean().optional().describe('Close through the Tempo precompile'),
+      orphaned: z.boolean().optional().describe('Close every recovered orphaned session'),
+      url: z.string().optional().describe('Override the stored resource URL'),
+      yes: z.boolean().optional().default(false).describe('Confirm closing every session'),
+    }),
+    output: z.object({
```
brendanjryan, line 1749: 👍

brendanjryan:
@codex review

brendanjryan (approved):

PR #725, opened 2026-07-29
refactor: remove legacy viem call shims

parvahuja:
## What

- remove legacy viem Tempo call-builder compatibility shims
- call token builders with their current client-first signatures
- call `dex.buy.call` with its current argument-only signature
- update focused tests to lock in those signatures

## Why

mppx now depends on a viem version with stable Tempo call-builder signatures. The runtime `Function.length` branches no longer provide compatibility and make the supported behavior harder to follow.

## Checks

- `VITE_TEMPO_NETWORK=none pnpm exec vp test --run src/tempo/client/Charge.test.ts src/tempo/internal/auto-swap.test.ts src/mcp/client/McpClient.integration.test.ts`
- `pnpm check:types`
- `pnpm check:ci`

### DIFF
--- a/.changeset/remove-viem-call-shims.md
+++ b/.changeset/remove-viem-call-shims.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Removed obsolete compatibility shims for legacy viem Tempo call builders.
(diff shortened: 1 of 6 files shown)


brendanjryan:
TY!

PR #726, opened 2026-07-29
fix: normalize Tempo address equality

parvahuja:
## What

- resolve Tempo-formatted and hex addresses before comparing them
- keep address comparison case-insensitive
- add focused coverage for hex, Tempo-formatted, and unequal addresses

## Why

The previous comparison lowercased the input strings directly, so equivalent hex and Tempo-formatted representations compared as different addresses.

## Checks

- `VITE_TEMPO_NETWORK=none pnpm exec vp test --run src/tempo/internal/address.test.ts src/tempo/internal/auto-swap.test.ts src/tempo/internal/fee-token.test.ts`
- `pnpm check:types`
- `pnpm check:ci`

### DIFF
--- a/src/tempo/internal/address.test.ts
+++ b/src/tempo/internal/address.test.ts
@@ -4,17 +4,39 @@ import { describe, expect, test } from 'vp/test'
 import { isEqual } from './address.js'
 
 const address = '0x742d35Cc6634C0532925a3b844Bc9e7595F2bD28'
+const otherAddress = '0x52908400098527886E0F7030069857D2E4169EE7'
 
 describe('isEqual', () => {
-  test('compares hex addresses case-insensitively', () => {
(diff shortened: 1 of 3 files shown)


src/tempo/internal/address.test.ts
```diff
@@ -0,0 +1,20 @@
+import { TempoAddress } from 'ox/tempo'
+import { describe, expect, test } from 'vp/test'
+
+import { isEqual } from './address.js'
+
+const address = '0x742d35Cc6634C0532925a3b844Bc9e7595F2bD28'
+
+describe('isEqual', () => {
+  test('compares hex addresses case-insensitively', () => {
```
brendanjryan: can we add additional cases here? I think thosu could be table driven + more exhaustive

brendanjryan (approved):

PR #731, opened 2026-07-30
[DNM] fix: evaluate settlement schedules during SSE streams

ksn-stripe:
## Summary

Metered SSE charges updated channel spend and units without evaluating the server settlement schedule. Ordinary paid HTTP requests already evaluated the same schedule after charging.

This change:
- evaluates settlementSchedule after each successful non-zero SSE charge commit through the session-created SSE transport
- keeps settlement conditional on the existing amount, unit, or interval thresholds via maybeSettleScheduled
- documents intervalMs as opportunistic: the first charge after the interval elapses triggers evaluation; no background timer is added
- exposes the shared post-commit callback through standalone Ws.serve, but does not claim automatic WebSocket schedule wiring because session() only constructs the SSE transport

## Tests

- Added a session integration regression that constructs session({ settlementSchedule, sse: true }), drains its real SSE transport, verifies no settlement before the amount threshold, and verifies the scheduled settlement and persisted boundary after crossing it.
- Added metered transport coverage for successful commits, no-op commits, and failed commits.
- Passed: focused node tests for Transports, Settlement, and Sse (55 tests).
- Passed: repository-wide lint and format check.
- Locally blocked: Session.test.ts collection and root typecheck require generated HTML modules. pnpm build:html fails because the local esbuild service is stopped; example typechecks also cannot resolve the unbuilt mppx package. CI should run the session regression and complete typechecks.

### DIFF
--- a/.changeset/metered-stream-settlement-schedule.md
+++ b/.changeset/metered-stream-settlement-schedule.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Added settlement schedule evaluation after successfully committed SSE charges in session handlers and exposed the same post-commit hook for standalone WebSocket metering.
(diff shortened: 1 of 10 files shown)


ksn-stripe:
Closing for now. Want to chat with the Tempo team about the edge case before proposing a fix.

PR #732, opened 2026-07-30
feat: add TAP and Web Bot Auth attestation

parvahuja:
## Why

MPP handles the HTTP 402 payment exchange, while services may also need authenticated agent identity and browse or payment intent. This binds those claims to selected signed HTTP components, resolves keys through application trust policy, and protects against replay; the application still decides what a valid signature authorizes.

## What

- Adds RFC 9421 primitives plus TAP and Web Bot Auth HTTPS-directory-profile signers and verifiers.
- Composes signers with `fetch`, sharing context within an attempt and refreshing it for retries.
- Exposes typed verification outcomes, injected key resolvers, and nonce storage. It does not include a key registry, directory service, or authorization middleware.


src/attestation/internal/HttpMessageSignature.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      signature:
+        'sig2=:RdNFx5Bj6au3YgAMQL/RzmUlZE8QZLIaXGRpw985hWnwPfMxT228NMk6ehRS1PSl4e8PhbNZACSanGdhEwYCCg==:',
+    })
+
+    expect(await verifyVector(request, key)).toMatchObject({ status: 'verified' })
+  })
+
+  test('verifies the RSA-PSS SHA-512 Signature-Agent vector', async () => {
+    const spki = Buffer.from(
+      'MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAr4tmm3r20Wd/PbqvP1s2+QEtvpuRaV8Yq40gjUR8y2Rjxa6dpG2GXHbPfvMs8ct+Lh1GH45x28Rw3Ry53mm+oAXjyQ86OnDkZ5N8lYbggD4O3w6M6pAvLkhk95AndTrifbIFPNU8PPMO7OyrFAHqgDsznjPFmTOtCEcN2Z1FpWgchwuYLPL+Wokqltd11nqqzi+bJ9cvSKADYdUAAN5WUtzdpiy6LbTgSxP7ociU4Tn0g5I6aDZJ7A8Lzo0KSyZYoA485mqcO0GVAdVw9lq4aOT9v6d+nb4bnNkQVklLQ3fVAvJm+xdDOp9LCNCN48V2pnDOkFV6+U9nV5oyc6XI2wIDAQAB',
```
brendanjryan, line 36: nice

src/attestation/internal/HttpMessageSignature.ts
```diff
@@ -0,0 +1,409 @@
...
+    id: id as string,
+    parameters,
+  }))
+  const created = parameters.get('created')
+  const expires = parameters.get('expires')
+  const algorithm = parameters.get('alg')
+  const keyId = parameters.get('keyid')
+  const nonce = parameters.get('nonce')
+  const tag = parameters.get('tag')
+  if (
```
brendanjryan, line 289: can we do this a diff way?

src/attestation/internal/HttpMessageSignature.ts
```diff
@@ -0,0 +1,409 @@
...
+    keyId,
+    label,
+    nonce,
+    parameters,
+    tag,
+  }
+}
+
+function algorithmFromKey(key: CryptoKey): SignatureAlgorithm {
+  if (key.algorithm.name === 'Ed25519') return Algorithms.ed25519
```
brendanjryan: can we make the algos constants? we should be able to make a constant map

src/attestation/internal/HttpMessageSignature.ts
```diff
@@ -0,0 +1,409 @@
...
+  }
+  if (component.id === Constants.components.path) {
+    if (component.parameters?.size) throw new Error('Unsupported parameters for "@path".')
+    return url.pathname
+  }
+  if (component.parameters?.size)
+    return componentDictionaryValue(request, component.id, component.parameters)
+  const value = request.headers.get(component.id)
+  if (value === null) throw new Error(`Missing signed HTTP header "${component.id}".`)
+  if (!/^[\x20-\x7e]*$/.test(value))
```
brendanjryan, line 360: ?

src/attestation/internal/HttpMessageSignature.ts
```diff
@@ -0,0 +1,409 @@
...
+}
+
+function hasExpectedTag(entry: unknown, tags: ReadonlySet<string>): boolean {
+  const value = entry as Item | InnerList
+  if (!isInnerList(value)) return false
+  const tag = value[1].get('tag')
+  return typeof tag === 'string' && tags.has(tag)
+}
+
+function serializeComponent(component: Component): string {
```
brendanjryan, line 402: we should check this file and see if there is a library -- IIRC cloudflare vendors one

src/attestation/Constants.ts
```diff
@@ -0,0 +1,15 @@
+/** RFC 9421 signature algorithms supported by request attestation. */
+export const Algorithms = {
+  /** EdDSA using the Ed25519 curve. */
+  ed25519: 'ed25519',
```
brendanjryan, line 4: use these in that other file

src/attestation/NonceStore.ts
```diff
@@ -0,0 +1,29 @@
+import type { MaybePromise } from '../internal/types.js'
+
+/** Atomically records signature nonces for replay protection. */
+export type Store = {
```
brendanjryan, line 9: how is this different than the other store interface?

src/attestation/internal/HttpMessageSignature.ts
parvahuja: added comment

src/attestation/internal/HttpMessageSignature.ts
parvahuja: i checked different libraries (including the cloudflare one) and don't think its a good fit - a lot of the stuff we have to do in here would just be wrapping those library methods anyway and we wouldnt really get much benefit

src/attestation/NonceStore.ts
parvahuja: i added surrounding comments to make it clearer, but basically the general store doesn't cleanly support TTL itself - we could combine / adapt the interfaces later if we wanted.

brendanjryan:
LGTM !

PR #750, opened 2026-08-03
feat(tempo): bound optimistic charge exposure

gakonst:
## Summary

- add atomic per-payer accounting for optimistic Tempo charge exposure
- reconcile pending receipts using the existing exact transfer, sender, and challenge-bound memo verification
- retain reverted optimistic payments as debt and fall back to confirmed settlement when debt plus pending charges reaches the configured cap
- support caller-owned scopes for aggregating economically interchangeable currencies with the same raw unit

## Motivation

An optimistic MPP charge can be broadcast successfully while later reverting onchain. The observed [SpendingLimitExceeded transaction](https://explore.tempo.xyz/tx/0x0770a9c00e85825986dc9e55fb34bfaf126fa994d75bd1460107005499a24fa5) produced an application success path before its receipt was known, which leaves unlimited optimistic mode open to unbounded unpaid service. Waiting for every confirmation closes that hole but adds settlement latency to every request.

This adds a bounded fast path:

```
debt + pending exposure + new charge <= cap => serve optimistically
otherwise                                    => wait for confirmation
```

Successful receipts release capacity. Explicit failures become durable debt and continue consuming the payer credit line. Confirmed fallback payments cover the current request; they do not silently forgive existing debt.

The cap is per payer and intentionally does not replace an application-level global breaker or Sybil policy.

## Validation

- `pnpm exec vp test --run --project node src/tempo/server/CreditBudget.test.ts` — 5 passed
- `pnpm exec vp test --run --project node src/tempo/server/Charge.test.ts` — 125 passed
- `pnpm check:types`
- `pnpm check:ci`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/.changeset/bounded-optimistic-credit.md
+++ b/.changeset/bounded-optimistic-credit.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Added bounded per-payer credit accounting for optimistic Tempo charges, with confirmed fallback when pending charges and reverted-payment debt reach the configured exposure limit.
--- a/src/tempo/server/Charge.test.ts
+++ b/src/tempo/server/Charge.test.ts
@@ -3048,6 +3048,115 @@ describe('tempo', () => {
     })
   })
 
+  describe('intent: charge; bounded optimistic credit', () => {
+    test('retains an under-cap charge as pending payee exposure', async () => {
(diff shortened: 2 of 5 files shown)


brendanjryan:
chatted offline and we are going to merge this into app logic

PR #751, opened 2026-08-03
feat: add server and method offer selection hooks

brendanjryan:
## Motivation

Servers need to vary composed payment offers using application state and method-specific request values, such as exposing Stripe only when a charge meets its minimum amount.

## Summary

- Added a typed async `selectOffers` hook to `Mppx.create()` for composed HTTP offers.
- Added a typed async method-level `canOffer` hook to every server method and intent constructor.
- Applied Stripe currency minimums before optional application-defined Stripe policies.
- Exposed immutable method descriptors and deeply immutable normalized request snapshots to selectors.
- Preserved request URL, method, and headers when upstream bodies are already consumed or locked.
- Applied filtering before challenges, including malformed and unmatched credential fallback paths.
- Preserved redemption of previously issued credentials against the full configured offer set.
- Refreshed transitive security overrides required by the repository audit gate.

## Key design considerations

- `Method.ComposableHooks<T>` defines the common constructor contract, so `canOffer` is not provider- or intent-specific.
- Both hooks only filter existing offers: `canOffer` removes its configured method offer and `selectOffers` removes offers from the remaining set.
- Method-level `canOffer` runs first. Top-level `selectOffers` receives only method-approved offers.
- Each offer contains `key`, an immutable `{ name, intent, alias? }` method descriptor, and a deeply immutable schema-normalized `request`.
- Incoming `Request` objects are cloned independently for each hook. If an upstream body is already consumed or locked, selectors receive a bodyless clone with the original URL, method, headers, and signal.
- Valid credential redemption bypasses filtering so an issued credential can still settle. Malformed or unmatched credentials are rejected through the currently eligible offer set.
- Static discovery metadata retains every configured candidate because request-time selection is dynamic.

Stripe can be limited to charges of at least $1 while Tempo remains available directly on the Stripe method:

```ts
const stripeCharge = stripe.charge({
  client: stripeClient,
  currency: "usd",
  decimals: 2,
  networkId: "internal",
  paymentMethodTypes: ["card"],
  canOffer({ request }) {
    return BigInt(request.amount) >= 100n
  },
})

const payments = Mppx.create({
  methods: [
    tempo.charge({ currency: USDC, recipient }),
    stripeCharge,
  ],
})
```

The same policy can be expressed centrally with `selectOffers`:

```ts
const stripeCharge = stripe.charge({
  client: stripeClient,
  currency: "usd",
  decimals: 2,
  networkId: "internal",
  paymentMethodTypes: ["card"],
})

const payments = Mppx.create({
  methods: [
    tempo.charge({ currency: USDC, recipient }),
    stripeCharge,
  ],
  selectOffers(offers) {
    return offers.filter(
      (offer) =>
        offer.method.name !== "stripe" ||
        BigInt(offer.request.amount) >= 100n,
    )
  },
})
```

Selection can also depend only on the incoming request:

```ts
const payments = Mppx.create({
  methods,
  selectOffers(offers, { request }) {
    const isBrowser = request.headers.get("Accept")?.includes("text/html")
    if (isBrowser) return offers
    return offers.filter((offer) => offer.method.name !== "stripe")
  },
})
```

Aliases distinguish multiple implementations with the same wire method and intent:

```ts
selectOffers(offers) {
  return offers.filter((offer) => offer.method.alias === "alternateSession")
}
```


src/internal/types.ts
```diff
@@ -2,6 +2,15 @@
 // https://twitter.com/mattpocockuk/status/1622730173446557697?s=20&t=NdpAcmEFXY01xkqU3KO0Mg
 export type Compute<type> = { [key in keyof type]: type[key] } & unknown
 
+/** Recursively makes every property of a value readonly. */
```
bensandler-stripe, line 5: typescript rocks

src/evm/server/Charge.ts
```diff
@@ -107,7 +108,10 @@ export declare namespace charge {
   type Parameters = NativeConfig
   type Native = Method.Server<typeof Methods.charge, Defaults>
 
-  type NativeConfig = BaseConfig & CurrencyConfig & RecipientConfig
+  type NativeConfig = BaseConfig &
+    CurrencyConfig &
+    RecipientConfig &
+    Method.ComposableHooks<typeof Methods.charge>
```
bensandler-stripe, line 114: Method.ComposableHooks is very cool! I will try to rebase my method-level onPaymentSuccess on this.

brendanjryan: 🙇

bensandler-stripe:
My agent had positive things to say (it's building on this for the updated `stripe` library)

> Feedback for Brendan: his PR is solid. Our onPaymentSuccess fits naturally into ComposableHooks. His canOffer with minimumChargeAmountByCurrency already handles what we were manually setting. Clean composition.

bensandler-stripe:
One minor thing: when rebasing my branch on this PR, I hit a minor dep issue. mppx's package.json declares "ox": "1.0.5" but mppx's lockfile pins 0.14.33. The dist gets built against 0.14.33 (which exports TempoAddress) but consumers that resolve ox@1.0.5 get VirtualAddress instead and blow up at runtime.

Edit: Just realized this was not introduced by your PR and is on main. Let me put up a fix.

PR #755, opened 2026-08-03
fix(deps): align package.json versions with pnpm workspace overrides

bensandler-stripe:
## Summary
- Dependabot PR #737 bumped `ox`, `typescript`, and `file-type` in `package.json` but the `pnpm-workspace.yaml` overrides continued pinning the old versions, causing a silent mismatch where declared deps differed from what was actually installed.
- Reverts `package.json` to match the workspace overrides until these pins can be lifted together with their dependents (e.g., `ox` is pinned for `viem@2.55.10` compatibility).

**We can alternatively get rid of the overrides in pnpm-workspace.yaml if preferred! Not sure if that would cause churn for existing MPP integrations?**

| Package | package.json (was) | Actually installed (override) |
|---------|-------------------|-------------------------------|
| ox | 1.0.5 | 0.14.33 |
| typescript | ~7.0.2 | ~5.9.3 |
| file-type | 22.0.1 | 21.3.2 |

## Test plan
- [x] `pnpm install` reports "Already up to date" (lockfile unchanged)
- [x] `pnpm run check:types` passes

### DIFF
--- a/.changeset/spicy-words-count.md
+++ b/.changeset/spicy-words-count.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Align package.json dependency versions with pnpm-workspace.yaml overrides.
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -48,6 +48,9 @@ jobs:
             echo "✓ No pnpm.overrides in package.json"
           fi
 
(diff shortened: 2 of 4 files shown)


brendanjryan:
Superseded by #762, which centralizes the pinned versions through a pnpm catalog and removes the need for a custom drift checker.

PR #756, opened 2026-08-03
test: verify override drift check catches existing mismatches

bensandler-stripe:
## Purpose

This draft PR deliberately keeps the mismatched package.json versions (ox 1.0.5, typescript ~7.0.2, file-type 22.0.1) while adding the new `scripts/check-override-drift.ts` CI check.

**This PR is expected to FAIL CI** at the "Ensure package.json versions match workspace overrides" step, demonstrating that the check would have caught PR #737's drift before merge.

See companion PR #755 for the actual fix.

Do not merge — close after CI runs.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -48,6 +48,9 @@ jobs:
             echo "✓ No pnpm.overrides in package.json"
           fi
 
+      - name: Ensure package.json versions match workspace overrides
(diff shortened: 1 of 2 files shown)


bensandler-stripe:
CI failed as expected at the 'Ensure package.json versions match workspace overrides' step, confirming the drift check works. See #755 for the actual fix.

PR #768, opened 2026-08-06
feat(server): multi-method intent handler with implicit compose

bensandler-stripe:
## Motivation

When multiple methods share an intent (e.g. tempo/charge + stripe/charge), the shorthand `mppx.charge(...)` previously did not work; you had to instead use `mppx.compose(["tempo/charge", ...], ["stripe/charge", ...])`. This makes the common multi-method case verbose.

## Summary

- Replace `UniqueIntentHandlers` (only exposed shorthand for single-method intents) with `IntentHandlers` (exposes shorthand for all intents)
- Single-method intent: direct handler (unchanged behavior)
- Multi-method intent: implicit compose wrapper that fans out the same options to all matching methods
- Uses `composeHandlers` with `selectOffers` so the implicit compose respects server-level offer filtering
- Add `MultiMethodFn` type that accepts the intersection of all matching methods' options. If methods have incompatible required fields that can't be satisfied by a single options object, the call won't typecheck.

```ts
// Before: required explicit compose
const result = await mppx.compose(
  ['tempo/charge', { amount: '0.50' }],
  ['stripe/charge', { amount: '0.50', currency: 'usd', decimals: 2 }],
)(request)

// After: shorthand works for shared intents
const result = await mppx.charge({ amount: '0.50' })(request)
```

### DIFF
--- a/.changeset/multi-method-intent-handler.md
+++ b/.changeset/multi-method-intent-handler.md
@@ -1,0 +1,5 @@
+---
+'mppx': minor
+---
+
+Multi-method intent handler: `.charge()` now works when multiple methods share an intent, internally composing all matching methods into one handler that respects `selectOffers`.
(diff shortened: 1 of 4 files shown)


brendanjryan:
LGTM -- going to merge over flake

PR #770, opened 2026-08-06
examples: add missing secretKey to charge, subscription, and multi-fetch servers

cdrn:
Three example servers call `Mppx.create` without a `secretKey` and with no `MPP_SECRET_KEY` source anywhere in the example, so they throw at module load:

```
Error: Missing secret key. Set the MPP_SECRET_KEY environment variable or pass `secretKey` to Mppx.create().
```

- `examples/charge/src/server.ts`
- `examples/subscription/src/server.ts`
- `examples/session/multi-fetch/src/server.ts`

The sse and ws session examples already handle this with a demo-key fallback plus `MPP_SECRET_KEY` override, so this just applies the same pattern (and the same explanatory comment) to the other three. Verified by running the multi-fetch example before and after: the original fails on first request with the error above, the fixed one serves. `pnpm check:types:examples` passes.

No changeset since this only touches examples, happy to add one if you'd prefer.

FWIW the docs examples have the same gap - the pay-as-you-go and one-time-payments session snippets on mpp.dev / tempo.xyz omit `secretKey` from `Mppx.create` (the one-time ones use a `crypto.randomBytes` fallback, the session ones nothing at all), which is how I hit this in the first place (via tempoxyz/mpp-irl#10). Can open a PR against the docs repo too if that's useful.

### DIFF
--- a/examples/charge/src/server.ts
+++ b/examples/charge/src/server.ts
@@ -6,7 +6,13 @@ import { Actions, Chain } from 'viem/tempo'
 const account = privateKeyToAccount(generatePrivateKey())
 const currency = '0x20c0000000000000000000000000000000000000' as const // pathUSD
 
+// `Mppx.create()` requires a secret key so challenge IDs can be verified
+// statelessly. The example ships with a default demo key so it works
+// out of the box, but still allows override via `MPP_SECRET_KEY`.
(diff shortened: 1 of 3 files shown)


brendanjryan:
ty!

PR #764, opened 2026-08-05
Add stripe.create() higher-level interface

bensandler-stripe:
## Motivation

Setting up MPP integrations with mppx on top of Stripe requires ~100 lines of boilerplate: resolving deposit addresses, configuring each method with test/live tokens, recording crypto payments as PaymentIntents, and wiring up the right amount conversions. 

This PR adds a higher level interface that encapsulates much more of this complexity, allowing integrations to write much less code (example PRs: https://github.com/stripe-wpp/machine-payments-examples/pull/48 https://github.com/sandlerben/starlink-payment-api/pull/6).

### Goals (h/t @brendanjryan)
- Single interface to configure stripe + MPP (users should only need to config once) 
- Users can configure manually + choose payment methods if they want
- Users can compose stripe methods with non-stripe methods (e.g. cards on Stripe, solana off Stripe)
- Automatic configuration of new payment methods / destinations supported on stripe
- Support for both x402 and MPP
- Support methods supported by Stripe but which run through external MPP sdks (e.g. Solana)

## Summary

- Add `stripe.create()` factory that configures Stripe machine payments methods from `{ client, networkId, livemode }`
- `defaultMethods()` returns pre-configured tempo + SPT methods; sync with static addresses, async with API resolution (cached)
- `.additional()` for EVM (base), tempo sessions, and custom rails (solana); custom rail keys restricted to `stripe.Network` at compile time
- Crypto payments automatically recorded as Stripe PaymentIntents via per-method `onPaymentSuccess` hooks
- Normalize charge amounts so `.charge()` works with human-readable dollar amounts across all built-in methods
- Declarative builders with exhaustive `Record` types for defaults and additional methods
- `ConnectConfig` bundles `stripeAccount` with PI-specific params

## Examples

Async (deposit addresses resolved from Stripe API):

```ts
import { Mppx, stripe } from 'mppx/server'

const stripeMachinePayments = stripe.create({
  client: stripeClient,
  networkId: process.env.STRIPE_PROFILE_ID!,
  livemode: !process.env.STRIPE_SECRET_KEY!.includes('_test_'),
})

const mppx = Mppx.create({
  methods: await stripeMachinePayments.defaultMethods(),
  secretKey: mppSecretKey,
})
```

Sync (static deposit addresses, no API call):

```ts
const stripeMachinePayments = stripe.create({
  client: stripeClient,
  networkId: process.env.STRIPE_PROFILE_ID!,
[...]

PR #776, opened 2026-08-09
feat(stripe): reject sub-cent amounts and add new tests

bensandler-stripe:
A few minor follow-ups from #764:

- Add `canOffer` for Stripe-wrapped crypto payment methods that filter out below 1 cent.
- Forward `metadata` to `PaymentIntent` creation for all methods
- Add tests ensuring that Stripe and non-Stripe methods can compose together.


brendanjryan (approved):
LGTM

PR #784, opened 2026-08-12
feat(tempo): support machine-token payments

parvahuja:
## Why

MPP clients should prefer an MPP-owned machine token when a Tempo payment method enables it, while merchants continue to price and settle in their normal currency. The same interface should extend to other Tempo methods without exposing token-specific details.

## What

- add `machineTokenEnabled` to `tempo.charge` and the global `tempo` constructor
- resolve one canonical first-party token-to-currency swap route for client construction and server validation
- try the machine-token route before existing funding behavior
- support atomic client-broadcast push and server-broadcast pull, including local and hosted fee sponsorship

### DIFF
--- a/src/tempo/server/Charge.ts
+++ b/src/tempo/server/Charge.ts
@@ -1,7 +1,9 @@
 import * as SignatureEnvelope from 'ox/tempo/SignatureEnvelope'
 import {
+  createClient,
   decodeFunctionData,
   formatUnits,
+  http,
   keccak256,
   parseEventLogs,
   type TransactionReceipt,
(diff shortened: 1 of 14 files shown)


src/tempo/server/Charge.ts
```diff
@@ -434,6 +468,10 @@ export function charge<const parameters extends charge.Parameters>(
       })()
       if (client.chain?.id !== chainId)
         throw new Error(`Client not configured with chainId ${chainId}.`)
+      if (request.machineTokenEnabled && relay)
+        throw new Error('`machineTokenEnabled` is not supported with a Tempo relay.')
```
brendanjryan, line 472: why not? just controlling scope for now?

src/tempo/server/Charge.ts
```diff
@@ -804,6 +849,9 @@ export declare namespace charge {
      * relay broadcasts pull credentials, while it recognizes a push
      * credential as already broadcast and returns its receipt without sending
      * it again.
+     *
+     * This option cannot be combined with `machineTokenEnabled` until the relay
+     * supports the first-party settlement route.
```
brendanjryan, line 854: 👍

PR #785, opened 2026-08-13
fix: close sessions with stale receipt spend

brendanjryan:
## Motivation

Persistent Tempo sessions can retain stale receipt spend after the server has charged content. Cooperative close then signs below server-recorded spend and is rejected.

## Summary

- Close HTTP/SSE fallback sessions at the highest locally authorized cumulative amount.
- Prefer refreshed server snapshot authorization when available.
- Accept close receipts whose captured spend is bounded by the signed close voucher.
- Add regression coverage for stale persisted spend and durable CLI accounting.

## Key design considerations

- Never signs above the client’s existing cumulative authorization.
- Preserves close-ready and socket delivery estimates for streaming sessions.
- Keeps snapshot validation bounded by local voucher state.

### DIFF
--- a/src/tempo/session/client/Runtime.ts
+++ b/src/tempo/session/client/Runtime.ts
@@ -657,7 +657,7 @@ export function isExpectedCloseReceipt(parameters: ExpectedCloseReceiptParameter
     receipt.challengeId === challengeId &&
     receipt.channelId === channelId &&
     receipt.acceptedCumulative === expectedCloseAmount &&
-    receipt.spent === expectedCloseAmount
+    BigInt(receipt.spent) <= BigInt(expectedCloseAmount)
   )
 }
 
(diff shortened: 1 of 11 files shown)


src/tempo/session/client/Runtime.ts
```diff
@@ -842,7 +842,7 @@ export function closedStateFromReceipt(receipt: SessionReceipt, entry: ChannelEn
  * Priority:
  * 1. Matching close-ready receipt spend.
  * 2. Matching socket delivery estimate (`deliveredChunks * tickCost`) clamped by cumulative.
- * 3. Latest receipt-tracked spend for HTTP/SSE.
+ * 3. Highest local cumulative authorization for HTTP/SSE.
```
brendanjryan: needs to be high watermark

PR #795, opened 2026-08-14
feat(store): native tryClaim fast path for redis and upstash via setNx

EfeDurmaz16:
## What

Adds an optional `setNx(key, value, expires) => Promise<boolean>` to `redis.Parameters` and `upstash.Parameters`. When a caller provides it, the adapter exposes a native `Store.tryClaim` that maps to a single `SET … PXAT <expires> NX` (upstash: `set(key, value, { nx: true, pxat: expires })`).

This extends the exact extension point #765 created: today only `memory()` has a native `tryClaim`; `redis()` / `upstash()` / `cloudflare()` fall back to the `update`-based read-modify-write.

## Why

The value is not just performance, and it is not a bug fix (nothing regressed — these adapters only forward caller-supplied primitives). It is about which primitive a caller must supply to get an atomic replay claim:

- **Upstash REST — capability gap.** Upstash REST [does not support `WATCH`](https://upstash.com/docs/redis/features/restapi) and is stateless per request, so a correct optimistic-lock `update` (the CAS the `tryClaim` fallback needs) is **not implementable** there. `SET … NX` is the only atomic first-use primitive. `@upstash/redis` `set` supports both [`nx` and `pxat`](https://upstash.com/docs/redis/sdks/ts/commands/string/set).
- **Redis (ioredis / node-redis) — 1-RTT fast path.** A caller *can* write a correct WATCH/MULTI `update`; `SET NX PXAT` collapses that loop into one round-trip.

`cloudflare()` is intentionally left out: Workers KV is eventually consistent, last-write-wins, with [no conditional/atomic write](https://developers.cloudflare.com/kv/api/write-key-value-pairs/), so it cannot back an atomic set-if-absent. A Worker that needs atomic claims should use `Store.upstash()` / `Store.redis()` (Upstash REST works from Workers), or Cloudflare's own recommendation, Durable Objects.

## Details

- `expires` is an absolute Unix-ms timestamp, so it maps to `PXAT` / `pxat`, never a relative `PX`/TTL. The marker value is a placeholder — the claim is key-existence under the TTL, and the NX path never reads it back (noted in a comment).
- Additive and backward-compatible: existing callers passing `{ get, set, del, update? }` are unaffected. The key-prefix wrapper already forwards a native `tryClaim`, so prefixed stores inherit it.
- Both adapters share one code path (`wrapSetNx`).

Example callers:

```ts
// ioredis / Valkey
Store.redis({ get, set, del,
  setNx: async (k, v, expires) => (await client.set(k, v, 'PXAT', expires, 'NX')) === 'OK' })

// @upstash/redis
Store.upstash({ get, set, del,
  setNx: async (k, v, expires) => (await redis.set(k, v, { nx: true, pxat: expires })) === 'OK' })
```

## Test

Added runtime tests (redis and upstash use the native `setNx` path over the `update` fallback; the fast path is forwarded through a key prefix). `Store.test.ts` passes 41/41 and `tsgo -b` (incl. `Store.test-d.ts`) is clean. I could not run the full suite locally (its global setup needs a Tempo localnet RPC unrelated to this change).

## Open question

Right now `setNx` accelerates `tryClaim` on a store that is already an `AtomicStore` (provides `update`). Would you also want providing `setNx` alone to brand the store as claim-capable, so a caller who only needs replay protection does not have to implement a full `update` CAS? That touches the `AtomicActions` type model, so I left it out — happy to follow up whichever way you prefer.

### DIFF
--- a/.changeset/redis-upstash-setnx-tryclaim.md
+++ b/.changeset/redis-upstash-setnx-tryclaim.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Added a native `Store.tryClaim` fast path to the `redis` and `upstash` adapters through an optional `setNx` primitive. When provided, a replay claim is a single atomic set-if-absent with an absolute expiry instead of the update-based read-modify-write fallback.
--- a/src/Store.test.ts
+++ b/src/Store.test.ts
@@ -237,6 +237,66 @@ describe('tryClaim', () => {
 
     expect(await Store.tryClaim(store, 'k', Date.now() + 60_000)).toBe(false)
   })
+
+  // A fake set-if-absent honoring an absolute-ms expiry, shared by the
+  // adapter fast-path tests.
+  function fakeSetNx(seen?: string[]) {
+    const claimed = new Map<string, number>()
+    return async (key: string, _value: string, expires: number) => {
+      seen?.push(key)
+      const existing = claimed.get(key)
+      if (existing !== undefined && existing > Date.now()) return false
+      claimed.set(key, expires)
+      return true
+    }
+  }
+
+  test('redis uses a native setNx fast path over the update fallback', async () => {
+    const kv = fakeStringKv()
+    let updateCalls = 0
+    const store = Store.redis({
+      get: kv.get,
+      set: kv.put,
+      del: kv.delete,
+      update: (key, fn) => {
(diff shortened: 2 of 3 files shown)


brendanjryan:
Superseded by #799.

PR #798, opened 2026-08-14
examples: add Peer Cash merchant cash-out flow

ADWilkinson:
## Summary

- add a Base mainnet MPP merchant example that prepares Peer Cash plans from confirmed USDC revenue
- count each successful payment receipt once and reserve planned amounts before asynchronous preparation
- keep the operator endpoint on a separate localhost listener and return unsigned transactions only
- pin the `undici-types` transitive dependency to a provenance-attested compatible release so the workspace trust policy remains effective

## Why

MPP handles the incoming machine payment. Peer Cash handles a separate merchant operation after settlement: moving Base USDC from the merchant-controlled recipient wallet into a fiat cash-out order. The example keeps those boundaries explicit and does not treat Peer Cash as a payer-side payment method.

## Safety boundaries

- the public server receives payments into `MPPX_RECIPIENT`
- only successful EVM settlement hooks increase tracked revenue
- receipt references are deduplicated
- the cash-out planner binds to `127.0.0.1`
- the planner never accepts private keys, signs, or broadcasts
- payout methods that need a post-deposit access-policy transaction fail closed in this minimal example
- the README calls out the in-memory accounting boundary and the state that a production host must persist

## Validation

- 4 focused accounting and planner tests
- cold server smoke: health and admin status return 200; the paid route returns an MPP + x402-compatible Base USDC 402 challenge
- `pnpm --filter peer-cash check:types`
- `pnpm check:types`
- `pnpm check:types:html`
- `pnpm check:types:examples`
- `pnpm check:ci`
- `pnpm audit --ignore-registry-errors` (no known vulnerabilities)
- `pnpm install --frozen-lockfile`

No changeset: this adds an example without changing the published `mppx` package.

### DIFF
--- a/examples/peer-cash/README.md
+++ b/examples/peer-cash/README.md
@@ -1,0 +1,58 @@
+# MPP revenue to Peer Cash
+
+A merchant server that accepts Base USDC over MPP, counts successful settlements, and prepares
+an unsigned Peer Cash plan once confirmed revenue reaches a threshold.
+
+The public MPP route and the cash-out planner are separate listeners. The planner binds to
+`127.0.0.1` and never accepts private keys, signs, or broadcasts. The merchant wallet that receives
+the MPP revenue keeps custody and decides whether to submit the returned transactions.
+
+## Setup
+
+```bash
+npx gitpick wevm/mppx/examples/peer-cash
+pnpm i
+
+export MPPX_RECIPIENT=0xMerchantWallet
+export X402_FACILITATOR_URL=https://your-facilitator.example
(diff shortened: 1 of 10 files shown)


brendanjryan:
Thanks for adding! I think a base example is fine, but we would like to avoid adding vendor specific examples here (e.g. peer cash). You are free to host this example in your own repos / docs

ADWilkinson:
Understood, thanks for the quick direction. We will keep the seller-side Peer Cash example in Peer-owned repos and docs, and will not pursue a vendor-specific example here.

PR #800, opened 2026-08-14
feat(tempo): support unified machine-token payments

parvahuja:
## Why

Merchants should be able to accept machineUSD for Tempo charges and sessions without changing the token or recipient they advertise. Clients should use machineUSD when the merchant permits it and the payer can use it, then fall back to the advertised token. Each configured chain should use one unified machine-token deployment while existing direct Mercator sessions remain unchanged.

## What

- Add implicit machineUSD selection with direct-token fallback.
- Configure Tempo charges and sessions with the same unified `{ swap, token }` deployment on mainnet and Moderato, while keeping their protocol adapters separate.
- Keep the selected session rail sticky, validate router authorization, use PathUSD for machine-session fees, and preserve correct close and refund behavior.

### DIFF
--- a/src/cli/sessions/store.ts
+++ b/src/cli/sessions/store.ts
@@ -18,6 +18,7 @@ import {
   isTempoSessionChallenge,
   type TempoSessionChallenge,
 } from '../../tempo/session/client/Transports.js'
+import * as Channel from '../../tempo/session/precompile/Channel.js'
 import type { SessionReceipt } from '../../tempo/session/precompile/Protocol.js'
 import * as z from '../../zod.js'
 
@@ -161,6 +162,12 @@ const storedChannelSchema = z.object({
   escrow: z.address(),
   chainId: z.number(),
   opened: z.boolean(),
+  paymentScope: z.optional(
+    z.object({
+      payee: z.address(),
+      token: z.address(),
+    }),
+  ),
 })
 const accountSchema = z.object({
   name: z.optional(z.string()),
@@ -253,8 +260,8 @@ export function sessionScopeKey(scope: SessionScope): string {
 export function sessionScope(channel: ChannelEntry): SessionScope {
   return {
     payer: channel.descriptor.payer,
-    payee: channel.descriptor.payee,
-    token: channel.descriptor.token,
+    payee: channel.paymentScope?.payee ?? channel.descriptor.payee,
+    token: channel.paymentScope?.token ?? channel.descriptor.token,
     escrow: channel.escrow,
     chainId: channel.chainId,
   }
@@ -321,9 +328,8 @@ export function createSessionRegistry(options: CreateSessionRegistryOptions = {}
     const previousValue = await readJson(file)
     const previous =
       previousValue === undefined ? undefined : parseStoredSession(previousValue, file)
-    if (previous) assertSameSession(previous, input, file)
-
     const challenge = parseSessionChallenge(input.challenge, file)
+    if (previous) assertSameSession(previous, input, file)
     assertChallengeMatchesChannel(challenge, channel, file)
     const receipt = input.receipt
       ? sanitizeReceipt(input.receipt, channel.channelId, file)
@@ -743,6 +749,12 @@ function channelIdentity(channel: ChannelEntry): object {
     ),
     escrow: channel.escrow.toLowerCase(),
     chainId: channel.chainId,
+    paymentScope: channel.paymentScope
+      ? {
+          payee: channel.paymentScope.payee.toLowerCase(),
+          token: channel.paymentScope.token.toLowerCase(),
+        }
+      : undefined,
   }
 }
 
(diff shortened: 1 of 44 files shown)


.agents/friction-log/20260818134157-headless-server-imports/friction.md
```diff
@@ -0,0 +1,24 @@
+---
```
parvahuja: remove this file

.changeset/lightweight-server-entrypoints.md
```diff
@@ -2,4 +2,4 @@
 'mppx': patch
```
parvahuja, line 2: combine into a single changeset / dont modify old changesets

src/tempo/server/ChargeHtml.ts
```diff
@@ -0,0 +1,28 @@
+import type { NoExtraKeys } from '../../internal/types.js'
```
parvahuja: ? this is a smell what is the need for this / what does it do?

src/tempo/internal/defaults.ts
```diff
@@ -23,14 +23,21 @@ export const currency = {
 /** First-party machine-token swap deployments. */
 export const machineToken = {
   [chainId.mainnet]: {
-    swap: '0xC6D32f013E0fA3e83B63Dc680E99826761595732',
-    token: '0x20C0000000000000000000003793c39601711f19',
+    feeToken: tokens.pathUsd,
+    session: true,
```
parvahuja: we shouldnt need to say `session: true` it just ships with sessions - also is there a way to have feeToken just default to pathUSD like it does for other payment types? does it need to be explicit here?

parvahuja: ideally this is just `swap` and `token`

src/tempo/internal/machine-token.ts
```diff
@@ -61,11 +121,233 @@ function getDeployment(chainId: number | undefined) {
...
+  if (!deployment || !('session' in deployment) || deployment.session !== true) return undefined
+  return deployment
+}
+
 /** Returns whether a first-party machine-token route is deployed on a chain. */
 export function isSupported(chainId: number | undefined): boolean {
   return !!getDeployment(chainId)
 }
 
+/** Returns whether the first-party machine token supports TIP-1034 sessions on a chain. */
```
parvahuja: we should ship this feature as if sessions is always supported, not a flag.

src/tempo/internal/machine-token.ts
```diff
@@ -1,15 +1,22 @@
 import type * as Hex from 'ox/Hex'
```
parvahuja: overall this file seems overcooked do we need all of these methods?

src/tempo/internal/signature-envelope.ts
```diff
@@ -0,0 +1,34 @@
+import { SignatureEnvelope } from 'ox/tempo'
```
parvahuja, line 1: can we just use signature envelope from ox? what does that impl miss that we add here?

src/tempo/server/headless.ts
```diff
@@ -0,0 +1,3 @@
+/** Headless Tempo server methods without browser payment-page assets. */
```
parvahuja: why was all of this added - why do we have a version of each method with and without browser payment-page assets

src/tempo/session/client/ChannelOps.ts
```diff
@@ -263,7 +264,7 @@ export async function createOpenPayload(
     account,
     calls: [...(parameters.prefixCalls ?? []), { to: escrow, data: openData }],
     feePayer: parameters.feePayer,
-    feeToken: parameters.token,
+    feeToken: parameters.feeToken ?? parameters.token,
```
parvahuja, line 269: ah i see why we need feeToken explictly - keep it above

src/tempo/internal/signature-envelope.ts
parvahuja: okay i see the simplification in voucher we can keep this

src/tempo/session/server/CredentialVerification.ts
```diff
@@ -98,6 +104,98 @@ export function validateChannelDescriptor(
   }
 }
 
+/** Resolves the fee-token allow-list for sponsored session management transactions. */
+export function resolveCredentialAllowedFeeTokens(parameters: {
```
parvahuja: overall, would our implementation simplify if we just defaulted feeToken to the payment token, unless its machineUSD, in which case its defaulted to pathUSD? we seem to be carrying around the fee token in multiple places and its adding extra cruft in other places i.e here in credential verification

src/cli/sessions/store.ts
```diff
@@ -161,6 +162,12 @@ const storedChannelSchema = z.object({
   escrow: z.address(),
   chainId: z.number(),
   opened: z.boolean(),
+  paymentScope: z.optional(
```
parvahuja, line 165: this is for caching the client "logical" store even though the server challenge may request a difference currency - it is necessary.

PR #801, opened 2026-08-15
feat(tempo): support machine-token sessions without refunds

parvahuja:
## Why

Merchants should enable machine-token-funded sessions with one flag, and deferring partial refunds removes the extra router and refund signatures from session credentials.

## What

- Keep server challenges in the merchant normal currency and recipient, with only `machineTokenEnabled: true` added.
- Reshape opted-in client channels to verified machineUSD virtual routes and settle them through the combined router with ordinary reserve vouchers.
- Complete fee-sponsored open and top-up transactions with the server configured liquid fee token, including pathUSD-sponsored machineUSD channels.
- Allow full-spend close and reject partial close before broadcast until the reserve supports restricted-token refunds.
- Preserve ordinary session and charge behavior and configure the verified no-refund Moderato deployment.

### DIFF
--- a/.changeset/machine-sessions-route.md
+++ b/.changeset/machine-sessions-route.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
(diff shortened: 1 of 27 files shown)


parvahuja:
Superseded by #800, which retains refundable machine-token sessions, router-domain protections, and the corrected PathUSD sponsorship path.

PR #805, opened 2026-08-17
fix: support standard x402 client payments

brendanjryan:
## Motivation

Standard x402 V2 providers do not include the optional mppx route-binding extension. The client rejected those otherwise payable EIP-3009 challenges, including services indexed through the CDP Bazaar.

References [#623](https://github.com/wevm/mppx/pull/623)

## Summary

- Added standard x402 exact EIP-3009 signing with random authorization nonces
- Preserved derived nonces for mppx route-bound challenges
- Filtered unsupported and policy-rejected x402 offers before challenge selection
- Added helper and fetch-path regression coverage for bound, unbound, mixed-offer, and fail-closed cases

## Key design considerations

- Standard x402 fallback still requires resource information and existing network, currency, and amount policies
- Non-mppx extensions are preserved without being treated as route binding
- Permit2 and offers missing EIP-3009 domain metadata remain unsupported and are skipped before signing

### DIFF
--- a/.changeset/standard-x402-exact-client.md
+++ b/.changeset/standard-x402-exact-client.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Added client signing support for standard x402 exact EIP-3009 offers without requiring mppx route-binding extensions.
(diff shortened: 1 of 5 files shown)


brendanjryan:
cc @parvahuja

PR #806, opened 2026-08-17
fix: support official x402 resource servers

brendanjryan:
## Motivation

Ensure x402 compatibility is tested against an independent official implementation instead of only mppx components.

## Summary

- accepted standard x402 EIP-3009 challenges without the optional mppx route-binding extension
- added a pinned official x402 resource server and facilitator integration suite
- wired dedicated x402 projects into required CI with a digest-pinned local chain

## Key design considerations

- retained route-bound nonces whenever the mppx extension is advertised
- used the official facilitator engine behind real local HTTP endpoints because no official facilitator container is published
- kept the suite hermetic with deterministic accounts, a local token, and no external funds or secrets


brendanjryan:
cc @parvahuja

PR #813, opened 2026-08-18
fix: escape non-Latin-1 characters in challenge auth-params

rsproule:
### Problem

`Challenge.serialize` writes the charge `description` into the `WWW-Authenticate` header raw. Header values must be ByteStrings (code points ≤ 0xFF), so any description containing an em dash, smart quote, or emoji makes `Response` construction throw in fetch-compliant runtimes:

```
TypeError: Cannot convert argument to a ByteString because the character at index 858 has a value of 8212 which is greater than 255.
    at Object.respondChallenge (.../mppx.mjs:6295)
```

Minimal repro:

```ts
const challenge = Challenge.from({
  id: 'abc',
  realm: 'shop.example',
  method: 'stripe',
  intent: 'charge',
  request: { amount: '1400', currency: 'usd' },
  description: '1 × Classmatic — General Admission',
})
new Headers({ 'WWW-Authenticate': Challenge.serialize(challenge) }) // throws
```

We hit this in production: merchant product titles routinely contain em dashes, so any integrator passing a title through `charge({ description })` gets an opaque crash on the challenge response. (Existing serialize/deserialize tests pass because they never place the serialized value into a real `Headers`/`Response`.)

### Fix

`authParam` now escapes characters above Latin-1 as `\uXXXX` inside quoted strings, and the quoted-string parser decodes them. The escape is unambiguous with the existing convention: serializers always double raw backslashes, so a bare `\u` can only come from this encoding. Older parsers degrade to rendering the literal escape in the display-only description rather than failing.

- Round-trip covered in the existing `test.each` escape table (em dash, surrogate pairs + quotes).
- New test asserts serialized challenges are constructible as real header values.
- Fuzz suite passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/witty-doors-wave.md
+++ b/.changeset/witty-doors-wave.md
@@ -1,0 +1,5 @@
+---
+"mppx": patch
+---
+
+Escape characters above Latin-1 in challenge auth-params so serialized challenges are always valid HTTP header values. Previously a charge `description` containing an em dash, smart quote, or emoji made `Response` construction throw (`Cannot convert argument to a ByteString`) in fetch-compliant runtimes when the challenge was written to the `WWW-Authenticate` header.
--- a/src/Challenge.test.ts
+++ b/src/Challenge.test.ts
@@ -508,6 +508,16 @@ describe('serialize', () => {
       escaped: 'Ends with slash \\\\',
       label: 'trailing backslash',
     },
+    {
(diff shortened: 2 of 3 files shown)


brendanjryan:
LGTM!

PR #821, opened 2026-08-20
Add alternate payment credential headers

raubrey-stripe:
When using MPP in conjunction with traditional Auth schemes that use the `Authorization` header (like an API Key riding in `Authorization: Bearer sk_...`) there are conflicts - namely, `Authorization` header does not support multiple values.

I'm proposing a change to the MPP challenge that indicates in the `WWW-Authenticate` response header what request header the client should put the payment challenge into. 

**Proposed new `header` field in challenge**
```
  HTTP/1.1 402 Payment Required
  WWW-Authenticate: Payment id="pay_abc123",
    realm="api.example.com",
    method="tempo",
    intent="charge",
    // This guy says you should not use the default Authorization header, but instead use Payement-Authorization
    header="Payment-Authorization", 
    request="eyJhbW91bnQiOiIxMDAwIn0"
```

If `header` is not included, the challenge should be included in the `Authorization` header.

### DIFF
--- a/src/Challenge.ts
+++ b/src/Challenge.ts
@@ -26,6 +26,10 @@ export const Schema = z.object({
   digest: z.optional(z.string().check(z.regex(/^sha-256=/, 'Invalid digest format'))),
   /** Optional expiration timestamp (ISO 8601). */
   expires: z.optional(z.datetime()),
+  /** Optional HTTP field name to carry the payment credential. When omitted, uses Authorization. */
+  header: z.optional(
+    z.string().check(z.regex(/^[!#$%&'*+.^_`|~0-9A-Za-z-]+$/, 'Invalid HTTP header name')),
+  ),
   /** Unique challenge identifier (HMAC-bound). */
   id: z.string().check(z.minLength(1)),
   /** Intent type (e.g., "charge", "session"). */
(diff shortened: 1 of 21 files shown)


src/Challenge.ts
```diff
@@ -26,6 +26,10 @@ export const Schema = z.object({
   digest: z.optional(z.string().check(z.regex(/^sha-256=/, 'Invalid digest format'))),
   /** Optional expiration timestamp (ISO 8601). */
   expires: z.optional(z.datetime()),
+  /** Optional HTTP field name to carry the payment credential. Defaults to Authorization. */
```
brendanjryan: Why default to Authorization? isn't that the standard value?

IMO we should do 

`nil -> Authorization`

and have `Payment-Authorization` as a constant 

we can then control using ^ by adding a flag to the server, "requires auth" or something similar

raubrey-stripe:
@brendanjryan incorporated your feedback to get the the `nil --> Authorization` bits! I also have attempted to upstream this in the spec: https://github.com/tempoxyz/mpp-specs/pull/328

plans/auth-header.md
```diff
@@ -0,0 +1,7 @@
+# Alternate payment credential header
+
+Payment challenges may advertise `header="Payment-Authorization"`. Clients send the serialized `Payment` credential in that header, allowing `Authorization` to retain an ordinary Bearer or Basic credential. Omitted `header` parameters remain compatible with the existing `Authorization` behavior.
```
brendanjryan: can we remove this?

brendanjryan (approved):

brendanjryan (approved):

PR #825, opened 2026-08-21
feat: add deferred client payment preparation

brendanjryan:
## Motivation

Applications may need to inspect and authorize a selected payment before creating a credential, while preserving MPPx challenge selection and protocol-specific credential attachment.

## Summary

- Add `mppx.preparePayment(response, options?)`, returning a typed `PreparedPayment`.
- Expose the selected `challenge` and `method`, all supported `challenges`, memoized `createCredential(context?)`, and protocol-aware `setCredential(request, credential)`.
- Accept `options.request` when challenge extraction depends on the request/response pair, including MCP-over-HTTP.
- Keep `createCredential()` as the convenience path over the new primitive.
- Add a patch changeset.

## Key design considerations

- Preparation uses existing `Accept-Payment`, `orderChallenges`, transport, expiration, validation, and event behavior.
- Signing remains deferred until `createCredential()` and expiration is checked again at that point.
- The credential signs the immutable challenge exposed for inspection while transport-local provenance is retained for attachment.
- Consumed HTTP response bodies remain valid for header-based challenge preparation.
- `PreparedPayment` is immutable and runtime-only because it retains transport-local protocol state.
- Usage:

```ts
const request = { method: 'POST' }
const response = await mppx.rawFetch('/resource', request)
const payment = await mppx.preparePayment(response, {
  request,
  orderChallenges: (candidates) =>
    candidates.filter(({ challenge }) => approvedIds.has(challenge.id)),
})

inspect(payment.challenge)

const credential = await payment.createCredential()
const init = payment.setCredential(request, credential)
const result = await mppx.rawFetch('/resource', init)
```

### DIFF
--- a/src/client/Mppx.ts
+++ b/src/client/Mppx.ts
@@ -57,22 +57,23 @@ export type Mppx<
    *
    * @example
    * ```ts
-   * const response = await mppx.rawFetch('/resource')
-   * const payment = await mppx.preparePayment(response)
+   * const request = { method: 'POST' }
+   * const response = await mppx.rawFetch('/resource', request)
+   * const payment = await mppx.preparePayment(response, { request })
    * inspect(payment.challenge)
    * const credential = await payment.createCredential()
-   * const init = payment.setCredential({}, credential)
+   * const init = payment.setCredential(request, credential)
    * ```
    */
   preparePayment: (
     response: Transport.ResponseOf<transport>,
-    options?: preparePayment.Options<FlattenMethods<methods>> | undefined,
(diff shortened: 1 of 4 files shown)


src/client/Mppx.ts
```diff
@@ -189,76 +337,24 @@ export function create<
...
-            challenges,
-            error,
-            method: mi,
-            response,
-          }),
-        )
-        throw error
-      }
+      const prepared = await preparePayment(response, options)
+      return prepared.createCredential(context)
```
brendanjryan, line 350: this just gets delegated!

PR #844, opened 2026-08-26
docs: add contract-qualified preparePayment example

epistemedeus:
## Summary

- add a bounded example that uses the exported `preparePayment()` seam to inspect an immutable selected challenge before credential creation
- qualify one frozen HTTP request and a deliberately narrow OpenAPI 3.1 success contract before calling `createCredential()`
- attach the credential to the original request and stop without sending an authenticated request
- add a focused oracle test covering exact matches, hostile mismatches, expiry, preparation without signing, and concurrent credential creation

## Boundary

MPP binds the selected challenge fields. This example records buyer-side application evidence for the HTTP request and output contract. It does not claim that MPP cryptographically binds the HTTP method, URL, query, body, or OpenAPI contract.

The change is example, documentation, changeset, and focused test only. It does not modify MPPx core, exports, transport, or credential semantics.

## Validation

```text
VITE_TEMPO_NETWORK=none vp test src/client/prepared-payment.example.test.ts --project node
19 passed
```

An independent controller replay also exercised 19 matching, mismatch, expiry, and concurrency cases against the real exported `Mppx.create().preparePayment()` path. Every mismatch and preparation-only case kept credential creation at zero. The matching and concurrent paths created one credential and sent no authenticated request.

AI assistance: implementation used Grok Bot. I independently reviewed the exact tree, replayed the load-bearing tests, and signed this commit.

### DIFF
--- a/.changeset/contract-qualified-prepare.md
+++ b/.changeset/contract-qualified-prepare.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Added a contract-qualified preparePayment example that binds a request and OpenAPI 3.1 success contract before creating a credential.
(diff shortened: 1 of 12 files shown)


brendanjryan:
Thanks! 

We aren't actively adding new examples here, but will consider this if we want to expand on this in the future

PR #850, opened 2026-08-28
feat(cli): support x402 payment challenges

brendanjryan:
## Motivation

Supersedes #812 by allowing the CLI to pay x402-only endpoints directly.

## Summary

- Added `--protocol auto|mpp|x402` challenge selection.
- Added x402 EVM signing and `PAYMENT-SIGNATURE` request handling.
- Documented the option and covered protocol selection, asset filtering, and failure paths.

## Key design considerations

- Automatic mode retains MPP-first selection while degrading safely around malformed x402 co-offers.
- The selected challenge retains protocol provenance through credential creation.


safal207:
The redirect hardening here exposed a useful protocol-level invariant, so I turned it into a small protocol-neutral red/green conformance fixture in AgentProof/Astra:

- **red:** origin A redirects to B, B issues the accepted x402 challenge, and the same typed payment authorization is observed at both A and B → `PAYMENT_CREDENTIAL_ORIGIN_DIVERGENCE` + `CROSS_ORIGIN_CREDENTIAL_REUSE`;
- **green:** the credential-bearing retry goes directly to B, dispatch and consumer evidence carry the same typed authorization identity, and one settlement, receipt, delivery, and reconciliation agree.

Merged implementation and claim boundary: https://github.com/safal207/safal207-AgentProof-AI/pull/19

The red fixture deliberately contains no settlement event: it proves credential exposure, not historical theft, consumption, or duplicate payment.

One contract question: should the challenge origin be the normative credential recipient for every reusable x402 credential, or do you expect authenticated intermediary/delegate origins? If delegates are intended, what exact artifact should establish that authorization so a client or independent verifier does not infer it from redirects or proxy topology?

src/cli/plugins/x402.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/** Returns whether the CLI can create an x402 credential for this challenge. */
+function isSupportedChallenge(challenge: { request: Record<string, unknown> }): boolean {
+  if (!x402_ChallengeBrand.is(challenge)) return false
+  const request = challenge.request
+  const transfer = request.extra as Record<string, unknown> | undefined
+  const transferMethod = transfer?.assetTransferMethod
+  return (
+    request.scheme === 'exact' &&
```
parvahuja, line 65: codex loves these long type checks - likely something simpler can be done

src/cli/cli.ts
```diff
@@ -166,18 +172,128 @@ function formatPayment(payment: unknown): string {
     .join(' ')
 }
 
+/** Returns the asset identifier used by a native MPP or x402 challenge. */
+function offeredCurrency(challenge: Challenge.Challenge): unknown {
```
parvahuja, line 176: nit: this could just return a string and that would simplify the check on 187

parvahuja (approved):

PR #857, opened 2026-08-31
fix: preserve multi-rail challenges in framework adapters

bensandler-stripe:
- Framework adapters rebuild shorthand intent handlers once per payment method. When multiple methods share an intent, each iteration replaces the previous shorthand, so the final method silently shadows the core composed handler and earlier payment offers disappear.

- This change wraps the shorthand handler already composed by the core when it differs from the per-method handler. Unique shorthand, slash, and nested handlers retain their existing behavior, and adapter `compose` remains intentionally omitted.

- Without this change, the following route returns only the Stripe challenge even though `defaultMethods()` supplies both Tempo and Stripe charge methods:

```ts
import express from "express"
import { Mppx, stripe } from "mppx/express"

const stripeMachinePayments = stripe.create({
  client: stripeClient,
  depositAddresses: { tempo: depositAddress },
  livemode: false,
  networkId,
})
const mppx = Mppx.create({
  methods: stripeMachinePayments.defaultMethods(),
  secretKey,
})

const app = express()
app.get("/paid", mppx.charge({ amount: "0.50" }), handler)
```

- An unauthenticated request to `/paid` now returns challenges for both `tempo/charge` and `stripe/charge`.

### DIFF
--- a/.changeset/fresh-rails-compose.md
+++ b/.changeset/fresh-rails-compose.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
+Fixed framework adapters to preserve multi-method intent challenges.
--- a/src/middlewares/internal/mppx.test.ts
+++ b/src/middlewares/internal/mppx.test.ts
@@ -148,6 +148,21 @@ describe('wrap: nested handlers', () => {
     expect(nestedResult.options).toEqual(slashResult.options)
   })
 
+  test('shorthand intent composes methods that share an intent', async () => {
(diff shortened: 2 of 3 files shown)


brendanjryan:
good find

brendanjryan (approved):

brendanjryan:
@bensandler-stripe can you just sign commits?

bensandler-stripe:
ah sorry about that again brendan!

bensandler-stripe:
@brendanjryan ready to merge

brendanjryan:
🚢

PR #858, opened 2026-08-31
Additional Payment Intent Options

staubman-stripe:
Builds on https://github.com/wevm/mppx/pull/849 to add 3 additional payment intent options: receipt email, customer id, and tax hooks. Since these additional options are more opportunities for the transaction verification to PI to fail, also adds more validation and a fallback in the crypto path to retry without the paymentIntentOptions.


brendanjryan (approved):
LGTM!

PR #859, opened 2026-09-01
feat(client): validate paid application responses

epistemedeus:
## Summary

Add an optional caller-owned `validateResponse` hook to `Fetch.from` and `Mppx.create`, plus a detached `PaidResponse` helper.

The hook runs only after a paid retry has reached a final response and any method-specific recovery has completed. Throwing rejects the application output without another request, credential, or payment. The error retains the current response and the receipt observed on the exact paid retry.

When the hook is absent, existing fetch behavior is unchanged.

## Evidence boundaries

- a validator cannot substitute a foreign response, receipt, challenge, or credential by throwing a prebuilt error
- a method recovery wrapper cannot replace or erase the receipt observed on the paid retry
- a response that cannot be cloned fails closed before the validator runs
- invalid application output is not reported as `payment.failed` and does not retry payment
- seller-declared OpenAPI output remains discovery evidence; the caller owns runtime validation

## Verification

- 14 focused paid-response tests passed
- 194 related client tests passed; 15 existing Tempo-localnet cases require `localhost:18545` and fail with that service disabled
- lint passed with zero warnings and errors
- source and test TypeScript checks passed
- HTML build passed
- signed commit tree matches the remotely reviewed candidate tree exactly

Related specification context: tempoxyz/mpp-specs#326 documents seller-declared OpenAPI output contracts, while this PR covers the separate client runtime boundary after payment.

### DIFF
--- a/.changeset/caller-owned-paid-response.md
+++ b/.changeset/caller-owned-paid-response.md
@@ -1,0 +1,5 @@
+---
+'mppx': minor
+---
+
+Added an optional caller-owned `validateResponse` hook and detached `PaidResponse` view so paid application output can be rejected after settlement without retrying payment or dropping receipt evidence.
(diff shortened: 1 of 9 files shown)


brendanjryan:
Thanks for the PR! We don't immediately want to add new hooks here without strong justification. Will consider this in the future if we see more consistent need from users

PR #863, opened 2026-09-02
feat: add CLI payment extensions

brendanjryan:
## Motivation

Allow method-independent payment policy and funding to run after challenge selection and confirmation, before Mppx creates a credential.

## Summary

- Add an `Extension` namespace and ordered `extensions` CLI configuration.
- Run payment preparation for request and sign flows.
- Thread optional credential context through extensions and method plugins.
- Document the extension lifecycle and publish it as a patch change.

## Key design considerations

- Extensions are explicit opt-ins loaded from the selected Mppx config.
- Challenge selection, confirmation, credential creation, response handling, and output remain owned by Mppx.
- Extension rejection fails closed before signing.


src/cli/plugins/plugin.ts
```diff
@@ -39,8 +39,10 @@ export interface Plugin {
     methods: Method.AnyClient[]
     /** Optional context to pass to createCredential */
     credentialContext?: unknown
-    /** Override credential creation entirely (e.g., delegating to an external CLI) */
-    createCredential?: ((response: Response) => Promise<string>) | undefined
+    /** Override credential creation entirely (e.g., delegating to an external CLI). */
+    createCredential?:
```
parvahuja, line 43: we have `credentialContext` at the global level and then again in `createCredential` -- likely can be consolidated in a later PR

parvahuja (approved):

PR #869, opened 2026-09-04
feat(tempo): restore MACH funding routes

parvahuja:
## Why

Allow mppx clients holding MACH to fund payments without merchants advertising or accepting MACH directly.

## What

- Carry machineTokenEnabled in authenticated method details while preserving the advertised settlement currency and recipient.
- Route MACH funding through the canonical swapper into the advertised settlement currency.
- Preserve verified direct-settlement fallback when a MACH route is unavailable.
- Expose immutable deployment metadata and add focused regression coverage.
- Refresh release-tooling dependencies and adjust the package-size budget for the restored runtime.


parvahuja:
@codex review

parvahuja:
@codex review

src/tempo/client/Charge.ts
```diff
@@ -130,31 +129,73 @@ export function charge(parameters: charge.Parameters = {}) {
...
           account: defaultAccount,
           chainId,
           operation: {
             kind: 'executeCalls',
-            ...(autoSwap ? {} : { calls: transferCalls }),
+            ...(calls ? { calls } : {}),
           },
         })) ?? defaultAccount
 
+      const machineTokenCandidate = machineTokenEnabled
```
brendanjryan, line 157: i don't love us just inlining a big block here with a conditional, but I think this is fine in the short term since we can collapse later

src/tempo/internal/machine-token-charge.ts
```diff
@@ -0,0 +1,171 @@
...
+  type Address,
+  type Client,
+} from 'viem'
+import { call, readContract } from 'viem/actions'
+import { Abis, Actions } from 'viem/tempo'
+
+import * as defaults from './defaults.js'
+
+const swapAbi = [
+  {
```
brendanjryan: is this a custom ABI? or can we pull this from the upstream viem types?

src/tempo/internal/machine-token-session.ts
```diff
@@ -0,0 +1,261 @@
...
+        token: parameters.token,
+      }) as never,
+    )
+    return (balance as bigint) >= parameters.amount
+  } catch {
+    return false
+  }
+}
+
+function authorizationDomain(router: Address, chainId: number) {
```
brendanjryan, line 125: we still don't support sessions on mainnet right? so this is just landing ahead of then?

README.md
```diff
@@ -83,6 +84,13 @@ Mppx.create({
 const res = await fetch('https://mpp.dev/api/ping/paid')
 ```
 
+Servers price payments in PathUSD, USDC.e, or another supported settlement token. When a server
```
brendanjryan: rm from readme

src/tempo/internal/machine-token-charge.ts
parvahuja: lts custom, but i can clean this up a bit of the redudancy here

src/tempo/internal/machine-token-session.ts
parvahuja: we actually do support session via the swapper based of our original design -- it gets around the TIP1034 "refund" issue by burning the and reminting back to the user

PR #872, opened 2026-09-06
feat(evm): separate credential validation from settlement

decofe:
Separates EVM payment validation from settlement using the existing `validateCredential` / `broadcastCredential` lifecycle. Validation checks terms, signature, payer identity, and the default facilitator's approval without settling; broadcast revalidates before settlement. Existing route handlers retain their validate-and-settle behavior.

Includes native and x402 lifecycle tests, and a patch changeset. The tests configure a signer up front, so this PR does not depend on deferred client signing or MCP composition. Custom settlers retain responsibility for chain-state checks and replay protection; validation is not a reservation. Fee-payer error normalization is unchanged.

Validation: 47 targeted tests passed; `pnpm check:ci`, `pnpm check:types`, and `git diff --check` passed. Payment tests use local signers and in-memory facilitators; live-chain settlement was not exercised.

Split from [#870](https://github.com/wevm/mppx/pull/870).

Prompted by: @brendanjryan


src/evm/server/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+```ts
+const options = { realm: 'api.example.com', request: { amount: '0.25' }, scope: 'job:123' }
+const validation = await server.validateCredential(credential, options)
+// Atomically claim the application operation using validation.source and the job identity.
+const receipt = await server.broadcastCredential(credential, options)
+// Persist the completed operation and receipt.reference.
+```
+
+Validation checks payment terms, expiry, signature, and payer identity. With the default
```
brendanjryan: remove

PR #873, opened 2026-09-06
feat(mcp): compose multiple payment offers

decofe:
Adds instance `compose` support for MCP SDK and raw JSON-RPC transports. Unpaid calls receive all configured offers in order; paid calls dispatch to one matching offer without trying another payment after verification fails. Error codes, request IDs, and receipt metadata are preserved.

Includes runtime and type tests, and a patch changeset. Incompatible method transports and HTTP-only offer hooks are rejected. Static `Mppx.compose` remains HTTP-only. This PR is independent of the EVM client and server changes; fee-payer error normalization is unchanged.

Validation: 129 targeted tests passed across MCP/HTTP transports and composition; `pnpm check:ci`, `pnpm check:types`, and `git diff --check` passed. Tests use an in-memory payment method without live settlement.

Split from [#870](https://github.com/wevm/mppx/pull/870).

Prompted by: @brendanjryan


src/mcp/server/README.md
```diff
@@ -0,0 +1,38 @@
+# MCP payment composition
+
```
brendanjryan: remove

brendanjryan:
cc @parvahuja

parvahuja (approved):

PR #880, opened 2026-09-09
fix: secure sponsored key authorizations

brendanjryan:
## Motivation

Prevent fee sponsors from installing unapproved access keys and bind subscription authorizations to a single server challenge.

## Summary

- Reject sponsored key authorizations unless explicitly allowed
- Bind subscription key signatures to the challenge ID witness
- Require matching witnesses during client and server verification

## Key design considerations

- Keep generic sponsorship locked down by default
- Allow key installation only in the verified subscription flow

### DIFF
--- a/.changeset/secure-payment-authorizations.md
+++ b/.changeset/secure-payment-authorizations.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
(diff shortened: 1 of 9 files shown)


brendanjryan:
Superseded by #881 and #882 so the sponsorship-policy and subscription-replay fixes can be reviewed independently.

PR #881, opened 2026-09-09
fix: make sponsored key authorizations configurable

brendanjryan:
## Motivation

Let fee sponsors reject access-key installation when their policy requires it without changing existing default behavior.

## Summary

- Allow fee-sponsored key authorizations by default
- Add an explicit policy opt-out for sponsors that reject access-key installation
- Update audited dependencies required by CI

## Key design considerations

- Preserve compatibility for existing sponsorship flows
- Apply rejection only when `allowKeyAuthorization` is explicitly false

### DIFF
--- a/.changeset/reject-sponsored-key-authorizations.md
+++ b/.changeset/reject-sponsored-key-authorizations.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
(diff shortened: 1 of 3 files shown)


brendanjryan:
cc @parvahuja

parvahuja (approved):

PR #887, opened 2026-09-10
fix: enforce primary and split recipient allowlists

brendanjryan:
## Motivation

Primary recipients and zero-amount proofs could bypass configured recipient allowlists.

## Summary

- Require the primary recipient and every split recipient to appear in the allowlist.
- Apply the same checks before signing zero-amount proofs.
- Clarify the option contract and add regression coverage.

## Key design considerations

- Compare addresses without case sensitivity and reject unexpected recipients before signing.

### DIFF
--- a/.changeset/charge-recipient-allowlist.md
+++ b/.changeset/charge-recipient-allowlist.md
@@ -1,0 +1,5 @@
+---
+'mppx': patch
+---
+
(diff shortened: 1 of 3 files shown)


brendanjryan:
@codex review

brendanjryan:
cc @parvahuja

parvahuja (approved):

PR #896, opened 2026-09-14
feat: add llms.txt validation to new suggested category

ksn-stripe:
We've heard from users that `llms.txt` files are useful for discovery, however this file is not strictly part of [the spec](https://paymentauth.org/draft-payment-discovery-01.html).

To include it in `mppx validate` without enforcing it as a strict requirement, this PR 
- Adds a new `suggested` category for validations 
- Adds the `llms.txt` validation to this category

The validation fetches `/llms.txt` from the server origin and checks that it is present and valid text.


brendanjryan (approved):

brendanjryan:
Thank you! one small lint issue

PR #909, opened 2026-09-19
fix: reject conflicting split-charge attribution memos

brendanjryan:
## Motivation

A split payment carrying attribution for multiple challenges can authorize another request after the first challenge's replay marker expires.

## Summary

- Reject conflicting challenge or realm attribution across matched payment transfers.
- Share the binding check between receipt logs and signed transaction calls.
- Add a patch changeset.

## Key design considerations

- Require at least one matching attribution memo while preserving plain transfers, non-MPP memos, and differing client fingerprints.
- Check matched payment transfers so unrelated transfers cannot establish challenge binding.
- Reject conflicting signed transactions before simulation or broadcast; preserve replay-store formats and retention.


brendanjryan:
cc @parvahuja

parvahuja (approved):

